PR #2, opened 2026-04-16
feat: bus `embed` composition + fix generate_if in module typecheck

DanielKellerM:
## Summary

- **Fix `generate_if` bus typecheck bug**: The driven-output check for bus ports used `default_param_map()` instead of port-site parameter overrides, causing false "output port is not driven" errors when using `BusAxi4<WRITE=0>` in `module` contexts. FSMs were unaffected because they use a different code path. Fix: build the param map from defaults + port overrides, matching the pattern already used in codegen.

- **Add `embed` bus composition**: New `embed prefix: BusName<PARAM=val>;` syntax inside bus bodies to compose buses from reusable sub-bus building blocks. Each embed flattens the referenced bus's signals with a name prefix. Supports param substitution, transitive embedding, two-pass resolve (file-order independent), and `.archi` round-trip.

- **Allow `embed` inside `generate_if` branches**: Enables conditional channel composition, e.g. a single `BusAxi4` with `generate_if READ` / `generate_if WRITE` branches that embed channel-level buses.

## Motivation

Defining AXI4 bus variants (read-only, write-only, full) required either:
1. One bus with `generate_if` — but this was broken for `module` contexts (only worked in `fsm`)
2. Three separate bus types — duplicating ~40 signal definitions 3x

With these changes, a single parameterized bus works correctly everywhere:

```
bus BusAxi4
  param READ:  const = 1;
  param WRITE: const = 1;

  generate_if READ
    embed ar: BusAxiAddr;
    embed r:  BusAxiRData;
  end generate_if

  generate_if WRITE
    embed aw: BusAxiAddr;
    embed w:  BusAxiWData;
    embed b:  BusAxiBresp;
  end generate_if
end bus BusAxi4
```

Used as `BusAxi4<WRITE=0>` for read-only, `BusAxi4<READ=0>` for write-only, or `BusAxi4` for full.

## Test plan

- [x] All 64 existing tests pass (12 unit + 52 integration)
- [x] New test: `bus_genif_bug/BusGenIf.arch` — `generate_if` with READ/WRITE in module ports
- [x] New test: `bus_embed/ReadMaster.arch` — basic embed composition + SV build + Verilator lint
- [x] New test: `bus_embed/ParamRemap.arch` — param name remapping through embed chain
- [x] New test: `bus_embed/TransitiveEmbed.arch` — three-level transitive embed
- [x] Verilator lint clean on all generated SV (zero warnings)
- [x] Separate compilation via `.archi` files round-trips correctly

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -62,6 +62,18 @@ pub struct BusDecl {
     pub params: Vec<ParamDecl>,
     pub signals: Vec<PortDecl>,  // direction = from initiator's perspective
     pub generates: Vec<BusGenerateIf>,  // conditional signal groups
+    pub embeds: Vec<BusEmbed>,  // embedded sub-buses
+    pub span: Span,
+}
+
+/// Embedded sub-bus inside a bus definition.
+/// `embed prefix: BusName<PARAM=val, ...>;`
+/// Flattens to `prefix_signal` for each signal in the referenced bus.
+#[derive(Debug, Clone)]
+pub struct BusEmbed {
+    pub prefix: Ident,
+    pub bus_name: Ident,
+    pub params: Vec<ParamAssign>,
     pub span: Span,
 }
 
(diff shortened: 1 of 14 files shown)


DanielKellerM:
Splitting into separate PRs: one for the typecheck bug fix, one for the embed feature.

PR #6, opened 2026-04-18
fix: sim codegen — handle package structs, struct-typed regs, and struct ports

nogate-ai:
## Summary

Fixes seven related bugs in `sim_codegen` that manifest only when a module uses struct-typed ports or regs (e.g. hwif_in/hwif_out packed structs generated by rdl2arch). The SV emission path (`arch build`) was unaffected — only `arch sim` / `--pybind` was broken for struct-bearing designs.

Discovered while wiring functional cocotb testbenches against SystemRDL-generated register blocks (a SystemRDL → ARCH → pybind sim flow that drives AXI4-Lite / APB4 transactions and checks register semantics).

## What changed

All in `src/sim_codegen.rs`:

1. **Package-scoped structs / enums were dropped from `VStructs.h`.** `gen_structs_file` only walked file-scope `Item::Enum`/`Item::Struct`; now descends into `Item::Package.{structs,enums}` as well (matches what `fn_items` collection already did a few lines earlier).

2. **Struct-typed port initializer was invalid C++.** Port init list emitted `hwif_in(0)` for a struct type. Now emits `hwif_in()` when the port type is `TypeExpr::Named`, matching the existing reg-init logic.

3. **VCD trace code bit-shifted struct ports.** `collect_trace_signals` now skips `TypeExpr::Named` ports, same rule already applied to reg decls.

4. **Pybind wrapper didn't register struct types.** `dut.hwif_in.field` raised `TypeError: Unregistered type : MyIpHwifIn`. Now emits `py::class_<Struct>(m, "Struct", py::module_local()).def(py::init<>()).def_readwrite(...)` for every struct in the compilation unit. `module_local()` lets multiple pybind modules sharing struct names (e.g. two cpuif variants of the same RDL design) coexist in one Python process.

5. **Struct-field widths in concat defaulted to 8.** `FieldAccess` width inference flattened the name and looked it up in a flat widths map; a miss returned `unwrap_or(8)`. So `{27'h0, ctrl_r.mode, ctrl_r.enable}` lowered to `mode << 8` instead of `mode << 1` in the C++ model — silently corrupting readback muxes for any register whose fields aren't byte-aligned. Fixed by:
   - Pre-populating widths with `"<reg>.<field>": <field_width>` entries for every struct-typed reg/port in the module (walking the referenced struct's `StructDecl`).
   - Making `FieldAccess` width inference look up the AST-level `<base>.<field>` key directly before falling back to the flat C++ name path.

6. **Struct-typed regs lost their reset value entirely.** `collect_stmt_assigns` only inserted `ExprKind::Ident` LHS into the assigned set. Struct regs (whose LHS is `FieldAccess` like `mixed_r.foo <= ...`) never reached the reset emission, so the `if (rst)` branch simply didn't touch them and they stayed at the C++ default-constructed zero. Fixed by also collecting the base of a `FieldAccess` LHS.

7. **`StructLiteral` lowered to a placeholder string.** The expression `reset rst => MixedReg { f1: 5, ... }` went to `_n_mixed_r = 0 /* struct literal */;` — type error or silently wrong depending on context. Now lowered via an immediately-invoked lambda:
   ```cpp
   [&](){ MixedReg _t; _t.f1 = 5; ...; return _t; }()
   ```
   The reset-init-value handler also routes `StructLiteral` through `cpp_expr` so real struct reset values land in the `if (rst) { ... }` branch.

## Test plan

- [ ] Existing arch-com test suite passes (no regressions expected — changes are only additive fallbacks and new code paths for struct ports/regs that previously errored or silently mis-compiled).
- [ ] Downstream: rdl2arch-generated register blocks now pass functional cocotb tests through `arch sim --pybind` for AXI4-Lite and APB4 cpuif variants, covering reset values, w1c/w1s/wclr/wot/wzc, rclr/rset, register arrays, regfile containers, hwif in/out mirroring. Every test body runs without changes against the Verilator path too; the two backends agree on register semantics.

### DIFF
--- a/src/sim_codegen.rs
+++ b/src/sim_codegen.rs
@@ -331,6 +331,36 @@ impl<'a> SimCodegen<'a> {
             .collect();
         let port_info_str = port_info_entries.join(",\n");
 
+        // Collect all struct types declared in the compilation unit (file-scope
+        // and inside packages). Register each with pybind so that struct-typed
+        // ports like `hwif_in: MyIpHwifIn` are reachable from Python as
+        // `dut.hwif_in.field`.
+        let mut all_structs: Vec<&StructDecl> = Vec::new();
+        for item in &self.source.items {
+            match item {
+                Item::Struct(s) => all_structs.push(s),
+                Item::Package(p) => { for s in &p.structs { all_structs.push(s); } }
+                _ => {}
+            }
+        }
+        let mut struct_bindings = String::new();
+        for s in &all_structs {
+            let sname = &s.name.name;
+            // `py::module_local()` scopes the struct type to this extension
+            // module so multiple pybind builds sharing struct names (e.g. two
+            // cpuif variants of the same design) can coexist in one process.
+            struct_bindings.push_str(&format!(
+                "    py::class_<{sname}>(m, \"{sname}\", py::module_local())\n        .def(py::init<>())\n"
+            ));
+            for f in &s.fields {
+                let fname = &f.name.name;
+                struct_bindings.push_str(&format!(
+                    "        .def_readwrite(\"{fname}\", &{sname}::{fname})\n"
+                ));
+            }
+            struct_bindings.push_str("        ;\n");
+        }
+
         let cpp = format!(
 r#"// Auto-generated pybind11 wrapper for {class}
 #include <pybind11/pybind11.h>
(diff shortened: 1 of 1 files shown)


nogate-ai:

PR #9, opened 2026-04-18
feat: allow reg and let declarations inside generate_for / generate_if

nogate-ai:
## Summary

Extends `GenItem` with `Reg(RegDecl)` and `Let(LetBinding)` variants. `generate_for` / `generate_if` bodies now accept the same `reg` and `let` forms as plain module bodies, so register-array-style patterns can be expressed idiomatically instead of requiring hand-unrolling.

## Before / after

Before:
```
generate_for i in 0..3
  port inp_i: in UInt<8>;
  port out_i: out UInt<8>;
end generate_for
reg r_0: UInt<8> reset rst => 0;
reg r_1: UInt<8> reset rst => 0;
reg r_2: UInt<8> reset rst => 0;   // hand-unrolled
reg r_3: UInt<8> reset rst => 0;
```

After:
```
generate_for i in 0..3
  port inp_i: in UInt<8>;
  port out_i: out UInt<8>;
  reg   r_i:   UInt<8> reset rst => 0;
  let   lout_i: UInt<8> = r_i;
end generate_for
```

## What changed

- `ast::GenItem` gains `Reg(RegDecl)` and `Let(LetBinding)` variants.
- Parser's `parse_gen_items_for` / `parse_gen_items_if` accept the `reg` and `let` tokens (delegating to the existing `parse_reg_decl` / `parse_let_binding`).
- Elaborator's `expand_generate_for` substitutes the loop variable into each item's name (via the existing `subst_name` suffix rewrite) and emits `ModuleBodyItem::RegDecl` / `LetBinding` per iteration. New helpers `subst_reg_decl`, `subst_let_binding`, `subst_reg_reset`.
- `subst_expr` now also rewrites identifier suffixes (`r_i` → `r_0`), so in-body expressions referencing peer unrolled names resolve correctly. Previously only declared names were renamed, not their uses.
- Reg / let items force elaboration-time expansion (same constraint as ports and threads); the loop range must be compile-time known.
- SV codegen added `unreachable!` arms for `Reg` / `Let` since they never survive elaboration.

## Test plan

- [ ] Existing `generate_for` / `generate_if` designs (ports, insts, threads, asserts only) behave unchanged — no new elaboration path unless `reg` or `let` is present.
- [ ] `reg r_i` inside `generate_for i in 0..N-1` produces N distinct registers with suffix-substituted names, resets, and init values.
- [ ] `let lout_i: T = r_i;` substitutes both declared name and `r_i` reference.
- [ ] Using `reg`/`let` with a non-const range yields a "must be compile-time constant" error (same as ports do today).

## Reviewer notes

Motivated by register-file-style generators (rdl2arch) that want to emit N identical registers without the caller hand-unrolling. Exercise code:

```arch
module GenRegLet
  port rst: in Reset<Sync>;
  port out_0: out UInt<8>;
  port out_1: out UInt<8>;
  port out_2: out UInt<8>;

  default seq on clk rising;

  generate_for i in 0..2
    reg r_i: UInt<8> reset rst => 0;
    let lout_i: UInt<8> = r_i;
  end generate_for

  seq
    r_0 <= 8'h11;
    r_1 <= 8'h22;
    r_2 <= 8'h33;
  end seq

  comb
    out_0 = lout_0; out_1 = lout_1; out_2 = lout_2;
  end comb
end module GenRegLet
```

`arch build` round-trips this to clean SV (`r_0..r_2`, `lout_0..lout_2`).

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -248,13 +248,17 @@ pub struct AssertDecl {
 
 // ── Generate ──────────────────────────────────────────────────────────────────
 
-/// An item inside a generate block: port, instance, or thread.
+/// An item inside a generate block: port, instance, thread, reg, or let.
+/// Reg and let are always expanded at elaboration time (there's no SV `generate
+/// for reg r: ...`), so using them forces the range to be compile-time known.
 #[derive(Debug, Clone)]
 pub enum GenItem {
     Port(PortDecl),
     Inst(InstDecl),
     Thread(ThreadBlock),
     Assert(AssertDecl),
+    Reg(RegDecl),
+    Let(LetBinding),
 }
 
 /// `generate for VAR in START..END ... end generate for VAR`
--- a/src/codegen.rs
+++ b/src/codegen.rs
@@ -2040,6 +2040,8 @@ impl<'a> Codegen<'a> {
                         GenItem::Inst(inst) => self.emit_inst(inst),
                         GenItem::Port(_) => unreachable!("port GenItems should have been lifted by elaboration"),
                         GenItem::Thread(_) => unreachable!("thread GenItems should have been lowered by elaboration"),
+                        GenItem::Reg(_) | GenItem::Let(_) => unreachable!(
+                            "reg/let GenItems should have been unrolled by elaboration"),
                         GenItem::Assert(_) => {
                             // SVA inside generate for: not yet supported in SV codegen (SVA needs static clock ref)
                         }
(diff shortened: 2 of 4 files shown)


nogate-ai:
Withdrawing. The suffix-unroll approach (r_i → r_0/r_1/r_2) produces ugly SV for non-trivial N, and my subst_expr change had a latent correctness issue — it would rewrite any identifier ending in _i (including peer names from outside the loop). Closing in favor of designing a proper Vec-based solution later.

PR #288, opened 2026-05-04
feat(params): post-name unpacked-array dim for SV passthrough params

nogate-ai:
## Summary

Native ARCH syntax for declaring a param with an SV unpacked-array post-name dimension:

```arch
param PMPRstCfg:        ibex_pkg::pmp_cfg_t [16]    = ibex_pkg::PmpCfgRst;
param PMPRstAddr[33:0]: const [16]                  = ibex_pkg::PmpAddrRst;
param PMPRstMsecCfg:    ibex_pkg::pmp_mseccfg_t     = ibex_pkg::PmpMseccfgRst;
```

Emits unchanged into SV:

```sv
parameter ibex_pkg::pmp_cfg_t PMPRstCfg [16] = ibex_pkg::PmpCfgRst,
parameter [33:0] PMPRstAddr [16] = ibex_pkg::PmpAddrRst,
parameter ibex_pkg::pmp_mseccfg_t PMPRstMsecCfg = ibex_pkg::PmpMseccfgRst,
```

Combines with the qualified-enum support from #286 — type qualifier (`ibex_pkg::pmp_cfg_t`) and default value (`ibex_pkg::PmpCfgRst`) flow through unchanged because cross-package `EnumVariant` codegen already preserves `pkg::Variant` verbatim when the qualifier doesn't name a known ARCH enum. The new piece is the post-name `[N]` unpacked dim, which arch-com had no syntax for.

## Why

Surfaced by arch-ibex C1 IbexCore swap. Upstream `ibex_top.sv` passes `ibex_pkg::pmp_cfg_t [PMP_MAX_REGIONS]` typed values to its `ibex_core` instance unconditionally. Under PMPEnable=0 the values are sunk, but they still need to type-match at SV elaboration. Without this fix, declaring `param PMPRstCfg: const = 0` trips Verilator's V3Width internal error trying to pattern-match the upstream struct default through a mismatched-typed parameter.

## Implementation

- **AST**: add `unpacked_size: Option<Expr>` to `ParamDecl`. Orthogonal to `ParamKind` so any kind (Const, WidthConst, EnumConst, Type, ConstVec) can carry an unpacked dim.
- **Parser**: after the post-colon kind block, optionally consume `[<expr>]`. Distinguished from the pre-colon `param NAME[hi:lo]: const` width qualifier by position (post-kind, pre-`=`).
- **Codegen**: append ` [<size>]` after the param name in `emit_param_decl`. All four match arms plumb the new `unpacked_str`.

## Test plan
- [x] `cargo test --release --test integration_test` → 316 passed (was 315; +1 new test, 0 regressions)
- [x] `test_unpacked_array_param_for_external_struct` covers all three shapes in one module (struct + dim, logic-packed + dim, scalar struct).
- [x] arch-ibex C1: with this fix landed, `IbexCore.arch` declares `PMPRstCfg`/`PMPRstAddr`/`PMPRstMsecCfg` as upstream-typed params, ibex_top can pass `pmp_cfg_t [N]` values through, and the V3Width internal error is resolved. (The remaining SoC-lint failures are port-shape mismatches in the implementer's hand-written `.archi` for `ibex_cs_registers`, which is C1 implementation work, not arch-com.)

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -230,6 +230,13 @@ pub struct ParamDecl {
     /// `local param` → emits SV `localparam` (not overridable at inst site)
     pub is_local: bool,
     pub span: Span,
+    /// Optional unpacked-array post-name dimension: `param NAME: T [N] = ...`.
+    /// Emits SV `parameter T NAME [N] = <default>` — the SV unpacked-array
+    /// param shape used by upstream Ibex for `pmp_cfg_t [PMP_MAX_REGIONS]`
+    /// and `logic [W:0] [N]` style declarations. arch-com forwards the
+    /// dimension verbatim and treats the param as opaque (no value
+    /// evaluation), since unpacked-array param values are SV-side only.
+    pub unpacked_size: Option<Expr>,
 }
 
 #[derive(Debug, Clone)]
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -254,6 +254,13 @@ pub enum ParamKind {
     /// Indexable as `NAME[i]` returning `T`. Inst-site overrides via
     /// `param NAME = {…};`.
     ConstVec(TypeExpr),
+    /// Logic-typed value const: `param NAME: UInt<W> = ...;` (or SInt /
+    /// Bool). Emits SV `parameter [W-1:0] NAME = <default>` —
+    /// the same shape as `WidthConst` but with type-first surface
+    /// syntax matching how ARCH writes ports / regs / wires
+    /// elsewhere. Used together with the post-name unpacked-dim
+    /// to express upstream-SV `parameter logic [W:0] NAME [N] = ...`.
+    Logic(TypeExpr),
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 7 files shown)


nogate-ai:
Addressed review feedback on the `[33:0]: const [16]` weirdness — replaced with a type-first surface form using `UInt<W> [N]` (consistent with how ARCH writes ports / regs / wires elsewhere):

```arch
// before
param PMPRstAddr[33:0]: const [16] = ibex_pkg::PmpAddrRst;

// after — clean native syntax
param PMPRstAddr: UInt<34> [16] = ibex_pkg::PmpAddrRst;
```

Both lower to the same SV `parameter [33:0] PMPRstAddr [16] = ibex_pkg::PmpAddrRst,`. The pre-colon `[hi:lo]: const` form still works for back-compat; `UInt<W>` is the recommended new form.

Added `ParamKind::Logic(TypeExpr)` variant; parser dispatches when peek is `UInt`/`SInt`/`Bool`. Codegen strips the `logic` keyword that `emit_port_type_str` prepends since SV `parameter` doesn't take `logic` in the same place `input/output` does. Test updated to exercise the new form.

316 tests pass. Ready for review.

PR #390, opened 2026-05-22
wait 0+ cycle until: 0-cycle Mealy thread wait + NIC-400 v2 perf checker

nogate-ai:
Answers \"are there pipeline stage bubbles, and can the testbench check for them\" — and then **fixes the bubble** with a new thread-wait form.

## The bubble (before)

Each `wait until X;` lowers to a Moore-style entry state: the thread samples X at posedge K, *then* advances to the do/until body — drives don't go out until cycle K+1. Two hops (master + slave) compounded into a 1-cycle AR and a 1-cycle R bubble, and the round-trip cycle through the resource-lock state machine capped throughput at ~0.32 t/c.

| Path | Spec §14.1 target | Moore (`wait until`) |
|---|---|---|
| AR forward | 0 cycles | 1 cycle |
| R return | 0 cycles | 1 cycle |
| AR throughput | 1 t/c | 0.32 t/c (8/25) |

## The fix: `wait 0+ cycle until X;`

A new Mealy-style wait form. When followed immediately by `do BODY until Y;` (with or without a wrapping `lock R … end lock`), the elaborator fuses the two into a single thread state whose comb drives are gated by X and whose transition guard is `X && Y`. When both conditions hold on the same posedge, the thread makes 0-cycle forward progress.

Standard `wait until` keeps its ≥1-cycle Moore semantics — the new form is opt-in.

## NIC-400 v2 perf (after)

| Path | Spec §14.1 target | Mealy (`wait 0+ cycle until`) |
|---|---|---|
| AR forward | 0 cycles | **0 cycles ✓** |
| R return | 0 cycles | **0 cycles ✓** |
| AR throughput | 1 t/c | **1.00 t/c ✓** (9/9) |

A per-cycle probe of `m.ar_valid` / `s.ar_valid` / `s.ar_ready` confirms that every posedge with valid && ready fires a handshake — no gaps. (The intermediate \"0.89 t/c (8/9)\" figure that appeared briefly in earlier commits was a TB counter off-by-one — the formula computed `total_cycles = last_seen_cycle + 1` even though `last_seen_cycle` was already 1-indexed. Fixed in the final commit.)

## Implementation

- **AST:** new `ThreadStmt::WaitUntilMealy(Expr, Span)`.
- **Parser:** literal `wait 0+ cycle until …;` form (the `0+` must be a single token — no space allowed).
- **Elaborate:** fusion in `partition_thread_body_impl` for both
  - `WaitUntilMealy + DoUntil`
  - `WaitUntilMealy + Lock(DoUntil)`
  
  Bus-read collection extended to the new variant so wait-condition signals participate in the comb fan-in.
- **NIC-400 v2 MasterPort/SlavePort:** switched to the Mealy form (single-token diff on each `wait until`).
- **Latency checker:** pinned to `MAX_LAT_AR=0`, `MAX_LAT_R=0`, throughput ≥ 1/1 — fails loudly if any future change re-introduces a bubble.

## Tests

- ✅ `test_wait_0plus_cycle_until_mealy_fusion` — end-to-end probe of the fusion on a minimal example (0-cycle latency, 1.0 t/c).
- ✅ `test_wait_0plus_requires_immediate_do_until` — diagnostic when the new form is not immediately followed by a `do..until`.
- ✅ All 455 existing unit tests pass.
- ✅ NIC-400 v2 smoke + parallel + hot-slave + OOO + latency tests all pass.
- ✅ Verilator `--lint-only --assert` clean.

## Inspecting the bubble

VCD waveform + a rising-edge probe script are checked in:

```bash
arch sim --wave tests/nic400/fab_latency.vcd \\
  tests/nic400/Nic400Fabric.arch tests/nic400/Nic400MasterPort.arch \\
  tests/nic400/Nic400SlavePort.arch tests/nic400/BusAxi4.arch \\
  --tb tests/nic400/tb_nic400_fabric_latency.cpp -o /tmp/fab_wave

tests/nic400/probe_ar_bubble.sh tests/nic400/fab_latency.vcd
```

## Test plan

- [x] All unit tests pass (455/455)
- [x] NIC-400 fabric latency checker passes at 0/0 cyc + 1.0 t/c
- [x] Verilator lint clean
- [x] No regression in monolithic v1 design (still passes its tests)

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -648,8 +648,22 @@ pub enum ThreadStmt {
     ForkTlmAssign(RegAssign),
     /// Join all outstanding nonblocking TLM issues in the current group.
     JoinAll(Span),
-    /// `wait until cond;`
+    /// `wait until cond;` — Moore-style wait. The thread sits in a
+    /// dedicated wait state with comb defaults active; at the next
+    /// posedge where `cond` is true, it advances to the next state.
+    /// Lower bound on wait time = 1 cycle even if `cond` is already
+    /// true when entering this stmt.
     WaitUntil(Expr, Span),
+    /// `wait 0+ cycle until cond;` — Mealy-style wait. Must be
+    /// immediately followed by `do ... until ...;`. The lowering fuses
+    /// the two into a single state whose comb drives are gated by the
+    /// mealy condition, and whose state-transition guard is the
+    /// conjunction `(wait cond) && (do-until exit cond)`. Result: when
+    /// both conditions are true on the same posedge, the thread
+    /// progresses with zero added cycles — eliminating the 1-cycle
+    /// "I'm in the entry wait state with defaults active" bubble that
+    /// the standard `wait until` form imposes.
+    WaitUntilMealy(Expr, Span),
     /// `wait N cycle;`
     WaitCycles(Expr, Span),
     /// `if cond ... elsif ... else ... end if`
--- a/src/elaborate.rs
+++ b/src/elaborate.rs
@@ -1117,6 +1117,9 @@ fn subst_thread_stmt(stmt: &ThreadStmt, var: &str, val: i64) -> ThreadStmt {
         ThreadStmt::WaitUntil(cond, sp) => {
             ThreadStmt::WaitUntil(subst_expr_names(cond.clone(), var, val), *sp)
         }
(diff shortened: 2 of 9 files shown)


nogate-ai:
Superseded by #392 — re-submitted on a fresh branch to retry CLA check.

PR #459, opened 2026-05-28
fix(typecheck): scope check_port_reg_timing to legacy `port reg` form only

nogate-ai:
## Summary

Closes the false positive on `Nic400WidthAdapter.arch` (`rd_axlen_r`, `wr_axlen_r`, `r_pack_r`, `r_resp_r`) that motivated this PR by scoping `check_port_reg_timing` to the **legacy `port reg`** keyword form only, using the existing `PortRegInfo::legacy_port_reg` flag.

This replaces the earlier `synthesized: bool` approach (see commit history). The earlier approach worked but added a new AST field for what is really a semantic distinction already encoded in `legacy_port_reg`.

## Rationale

The timing-mismatch warning exists to catch the **implicit 1-cycle latency** the legacy `port reg` keyword hides. The modern form `port: out pipe_reg<T, N>` makes the latency visible in the port signature — when a user writes `pipe_reg<T, 1>`, they have explicitly opted into the N-cycle latency, so the foot-gun the warning was designed to catch doesn't apply.

The parser at `src/parser.rs:1405` already distinguishes the two surfaces:

```rust
Some(PortRegInfo {
    // ...
    legacy_port_reg: pipe_reg_latency.is_none(),
})
```

Thread lowering at `src/elaborate.rs:2378` already emits its synthesized port regs with `legacy_port_reg: false` (matching the modern form):

```rust
Some(PortRegInfo {
    // ...
    legacy_port_reg: false,
})
```

Gating the timing check on `ri.legacy_port_reg` therefore silences in one line:

- the four false positives on `Nic400WidthAdapter.arch` — synthesized port regs are already non-legacy.
- user-written `port: out pipe_reg<T, N>` ports — semantically correct, the latency is declared.

…and continues to fire on legacy `port reg` outputs assigned inside state-dependent branches — the case the check exists to catch.

## Comparison to prior approach

| | `synthesized: bool` (old) | `legacy_port_reg` (this PR) |
|---|---|---|
| Files touched | 5 (ast, parser, elaborate, typecheck, tests) | 2 (typecheck, tests) |
| New AST field | yes | no |
| Silences thread-synthesized port regs | yes | yes |
| Silences user-written `pipe_reg<T, N>` | no | **yes** (correctly — explicit opt-in) |
| Semantic clarity of the gate | provenance marker | matches user-visible distinction |

## Warning message update

Tightened to mention both modern escapes:

> ``out_reg`` is a deprecated ``port reg`` output assigned inside a state-dependent branch — the implicit 1-cycle latency makes the output appear one cycle after the state that drives it. Migrate to either ``port out_reg: out T`` driven by ``comb`` (same-cycle output) or ``port out_reg: out pipe_reg<T, N>`` (explicit N-cycle registered output)

## Tests

Three regression tests pin the new gate:

- `test_check_port_reg_timing_fires_on_legacy_port_reg` — positive control. Legacy `port reg` assigned in state-dependent branch still fires.
- `test_check_port_reg_timing_skips_modern_pipe_reg` — **new**. User-written `pipe_reg<T, N>` is silent even when assigned in a state-dependent branch.
- `test_check_port_reg_timing_skips_synthesized_thread_lowered_regs` — same shape as the prior PR. Thread-lowering synthesized port regs are silent automatically.

## Verification

- [x] `arch check tests/nic400/Nic400WidthAdapter.arch tests/nic400/BusAxi4.arch` now prints `OK: no errors` with no warnings (was: 4 spurious warnings).
- [x] `cargo test --release` clean — 504/504 integration tests pass (3 new), 24 + 20 lib tests pass.
- [x] `cargo build --release` clean.

Pure internal typecheck pass scope tightening — no spec change, no AST surface change, no codegen change.

## Refs

- arch-com#447 §A — false positive surfaced
- arch-com#440 — where the warning first fired
- arch-com#463 — 2026-05-28 findings note; same spirit applied here (don't accept a paper-over flag when there's a tighter, already-encoded semantic gate)

### DIFF
--- a/src/typecheck.rs
+++ b/src/typecheck.rs
@@ -1809,14 +1809,24 @@ impl<'a> TypeChecker<'a> {
         }
     }
 
-    /// Warn when registered output ports are assigned inside state-dependent
-    /// if/elsif chains in seq blocks. This indicates the output has register
-    /// latency relative to the state that drives it — a common timing mismatch
-    /// with testbench models.
+    /// Warn when **deprecated** `port reg` outputs are assigned inside
+    /// state-dependent if/elsif chains in seq blocks. The motivation is the
+    /// implicit 1-cycle latency the legacy `port reg` form hides — a common
+    /// timing mismatch with testbench models that expect same-cycle outputs.
+    ///
+    /// Filter is `ri.legacy_port_reg` (not just `reg_info.is_some()`) so that:
+    ///   * user-written `port: out pipe_reg<T, N>` is silent — the user has
+    ///     explicitly opted into the N-cycle latency by writing the type, so
+    ///     the foot-gun the warning was designed to catch doesn't apply;
+    ///   * thread-lowering synthesizes port-regs with `legacy_port_reg: false`
+    ///     (see `src/elaborate.rs::lower_threads`), so internal artifacts the
+    ///     user can't act on are silent automatically — no separate
+    ///     `synthesized` flag needed.
     pub(crate) fn check_port_reg_timing(&mut self, m: &ModuleDecl) {
-        // Collect registered output port names.
+        // Collect names of *legacy-form* registered output ports only.
         let port_reg_names: HashSet<String> = m.ports.iter()
-            .filter(|p| p.reg_info.is_some() && p.direction == Direction::Out)
+            .filter(|p| p.direction == Direction::Out
+                     && p.reg_info.as_ref().is_some_and(|ri| ri.legacy_port_reg))
             .map(|p| p.name.name.clone())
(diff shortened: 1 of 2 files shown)


nogate-ai:
Force-pushed a smaller replacement: the gate is now `ri.legacy_port_reg` instead of a new `synthesized: bool` field. Same false positives silenced; user-written `port: out pipe_reg<T, N>` is also silenced as a bonus (semantically correct — the latency is declared explicitly). 5 files → 2 files, no new AST field. Full rationale + tests in the updated PR body.

PR #460, opened 2026-05-28
feat(sim_codegen): warn on non-priority mutex policy under --thread-sim

nogate-ai:
## Summary

Follow-up to arch-com#447 §3. The `--thread-sim` scheduler resolves mutex
contention as lowest-thread-index-wins, which is observationally identical to
`mutex<priority>`. Any non-priority policy (`round_robin`, `lru`, `weighted`,
`custom`) is silently downgraded — so equivalence checks against the default
FSM-lowered path (e.g. `arch sim --thread-sim both` on a fairness-sensitive
design) can false-PASS.

arch-com#455 fixed the doc half (`doc/thread_spec_section.md` §20.8.1,
`doc/thread_lowering_algorithm.md` §Liveness per policy). This PR fixes the
matching codegen-time half: one warning per affected resource, naming the
resource, declared policy, source span, and the doc cross-link.

Example output for `mutex<round_robin>`:

```
warning: --thread-sim ignores mutex policy `round_robin` on resource `shared_bus`
         (module `SharedBus`); the --thread-sim scheduler resolves contention as
         lowest-thread-index-wins (== priority). For fairness-sensitive
         behaviour, validate with the default `arch sim` or Verilator.
         See doc/thread_spec_section.md §20.8.1. (path.arch:265)
```

Example for `mutex<MyFn>` (custom hook is bypassed — stronger phrasing):

```
warning: --thread-sim ignores mutex policy `PickHigh` on resource `shared_lk`
         (module `M`); the custom hook `PickHigh` is COMPLETELY BYPASSED — the
         --thread-sim scheduler resolves contention as lowest-thread-index-wins
         (== priority). Validate custom-policy behaviour with the default
         `arch sim` or Verilator. See doc/thread_spec_section.md §20.8.1.
```

## Changes

- `sim_codegen::thread_sim::gen_module_thread_with_warnings(...)` — new entry
  that takes a `&mut Vec<CompileWarning>`. `gen_module_thread(...)` keeps its
  old signature and forwards to the new one with a throwaway sink, so existing
  callers (integration tests, anywhere else) keep working.
- Old hard error on `mutex<weighted<W>>` / `mutex<MyFn>` is now a warning —
  the design runs under `--thread-sim`, but the user sees a strongly-worded
  message. Custom policies get distinct phrasing because the user-defined
  `hook grant_select` is ignored entirely.
- `main.rs` routes the new warnings into the same stderr path that typecheck
  warnings use, with full `filename:offset` annotation. Plumbed only through
  the `--thread-sim parallel` codegen entry, so `arch sim` (default
  FSM-lowered path) and `arch build` stay silent. `mutex<priority>` produces
  no warning under any path.

## Tests

`tests/integration_test.rs`:

- `test_thread_sim_warns_on_non_priority_mutex_policy` — pins the warning
  shape for `mutex<round_robin>` (one resource = one warning; no spam).
- `test_thread_sim_no_warning_on_priority_mutex_policy` — absence property
  for the priority case.
- `test_thread_sim_warns_on_custom_mutex_policy_with_hook` — pins the
  BYPASSED phrasing and the hook function name appearing in the message body.

New `compile_to_thread_sim_collect_warnings(...)` helper mirrors the existing
`compile_to_thread_sim_h(...)` pipeline but returns the collected warnings
instead of the header text.

## Cross-refs

- arch-com#447 §3 — the gap this closes
- arch-com#455 — the doc half (already merged)
- arch-com#443 — the policy-gate relaxation that surfaced this gap

## Test plan

- [x] `cargo build --release` — clean
- [x] `cargo test --release --test integration_test thread_sim` — 7/7 pass (3 new + 4 existing)
- [x] `cargo test --release` — 500/500 pass, no regressions
- [x] Manual: `arch sim --thread-sim parallel resource_lock.arch` with
      `mutex<round_robin>` → warning fires
- [x] Manual: `arch sim resource_lock.arch` (default path) → no warning
- [x] Manual: `arch sim --thread-sim parallel custom_policy.arch` → BYPASSED warning


nogate-ai:
Flagging from the 2026-05-28 daily code-review pass ([#463](https://github.com/arch-hdl-lang/arch-com/pull/463)) — the warning-only approach is not the right closure for §3.

`mutex<round_robin>` (and `lru` / `weighted` / `custom`) is a documented user-facing fairness semantic. If `--thread-sim` doesn't deliver it, the scheduler is buggy, not the user's design. A codegen-time warning leaves the false-PASS surface live and shifts the burden to the user — they have to read every warn line before trusting an `arch sim --thread-sim both` equivalence check.

The original §3 finding in [#447](https://github.com/arch-hdl-lang/arch-com/pull/447) listed both options and called the scheduler fix "longer-term". The real fix is to teach the scheduler at `src/sim_codegen/thread_sim.rs:808-814` to honour the policy:

- per-resource `last_grant` index for `round_robin`
- LRU stack for `mutex<lru>`
- credit counter for `mutex<weighted<W>>`
- hook callback dispatch (fire `hook grant_select(...)` and use its one-hot result) for `mutex<MyFn>`

Once the scheduler honours the policy, the warning emitted by this PR stops carrying signal and should be removed. `doc/thread_spec_section.md` §20.8.1's fairness caveat can then disappear and `--thread-sim both` becomes a real cross-check again.

Suggestion: either close this PR or keep the warning as an interim transition step with a tracking issue for the scheduler fix. I'm happy to take a stab at the scheduler implementation in a separate PR if that's the right next step — though the credit/weighted policy and custom-hook calling convention are non-trivial design choices that probably want a quick alignment first.

Full context in [#463](https://github.com/arch-hdl-lang/arch-com/pull/463) Finding 2.

Issue #246, opened 2026-05-02
Whole-design comb-loop analysis (arch check is per-module)


nogate-ai:
## Problem

`arch check` validates each module in isolation. Combinational loops that arise only when modules are wired together (via shared external SV scaffolding, hybrid swap-in, etc.) are invisible.

In arch-ibex this surfaces as: `IbexAlu` and `IbexMultdivFast` are individually clean. When wired together inside upstream `ibex_ex_block.sv`, their `alu_adder_result_ext` ↔ multdiv operand path forms an intentional same-cycle adder-share loop. Verilator reports UNOPTFLAT; we suppress with `-Wno-UNOPTFLAT` and document the loop is functionally safe. But arch's own check has nothing to say either way — it didn't catch the loop, can't bless it, can't tell us whether the design has *additional* unintended cross-module loops.

This same gap also bears on the dead-skid trap (see #245): thread output ↔ consumer comb ↔ thread input is a cross-module loop arch can't see.

## Proposal

Two-phase:

1. **Whole-design comb graph**: extend `arch check` to walk the full module hierarchy (including upstream SV stubs declared via `extern module` or similar) and build a single comb-graph spanning all of them. Report SCCs.

2. **Loop classification**: every comb SCC is either (a) intended (e.g. shared adder, handshake), (b) racy, or (c) a bug. The user should be able to tag intended loops with a pragma so the check passes; everything else warns by default.

## Why this matters

- Catches cross-module bugs that the per-module check is structurally incapable of finding.
- Lets us turn `-Wno-UNOPTFLAT` back on in arch-ibex's verilator gate (currently a blanket suppression because arch can't distinguish the intentional ALU↔multdiv loop from anything else).
- Pairs with the dead-skid lint (#245) — both depend on cross-module comb-graph awareness.

## Notes

This was already on arch-ibex's "compiler feature gap" list (HANDOFF.md item, not in repo). Filing here so it has a tracker.

nogate-ai:
Closed after a 4-phase implementation arc.

**Phase 1 — MVP whole-design check** (#336): walk the inst hierarchy, build a comb graph keyed by `(inst_path, signal)`, run Tarjan SCC. Module-level `pragma comb_loops_allowed;` for blessing. Treats non-bodied children (extern / `.archi` stubs) as fully opaque.

[...]

PR #486, opened 2026-06-01
Cross-module dead-skid comb-feedback lint + thread-map overlay (issue #245)

### LINKED ISSUE #245: Static lint: thread reads same-cycle comb function of its own outputs ("dead-skid drop" trap)
## Problem

When a `thread` body has `wait` between active states, the lowering inserts a "dead-skid" sub-state with no comb assigns. During that cycle, any output ports the thread was driving fall to their default (typically 0). For most signals this is invisible — but for **cross-module combinational feedback loops**, the consumer sees the drives drop, recomputes a stale value, and the thread reads it back the next cycle.

This was the most expensive single class of bug in the arch-ibex Phase A work. The A6 `IbexMultdivFast` swap took 3 thread-implementation iterations partly because of this trap — the spec used `equal_to_zero_i` (the ALU's `(adder == 0)` output) to gate divide-by-zero short-circuit. When lowered to a thread, mid-walk dead-skid cycles dropped `alu_operand_*_o` to 0; the ALU then computed 0+0=0; `equal_to_zero_i` momentarily asserted; a separate dispatch branch waiting on it spuriously fired.

The workaround was to gate on `op_b_i == 0` directly (an upstream input) instead of `equal_to_zero_i` (an ALU-routed comb function of the thread's own outputs). The two are mathematically equivalent under the LSB-1-injection trick — but only one is robust to dead-skid.

## Proposal

Add a `arch check` lint that detects the pattern:

> A `thread` reads signal X. X is a same-cycle combinational function of one or more output ports the thread itself drives (transitively, across module boundaries).

Concretely, the check needs:

- per-thread set of "output ports driven from inside the thread"
- per-module comb-graph (already computed for `arch check` + UNOPTFLAT-style work)
- a transitive closure walk: starting from the thread's read-set, can we reach any of the thread's own outputs through purely-comb edges?

If yes → warn with the cycle path. The user can either (a) refactor to read the upstream input directly, or (b) suppress with a pragma if they know the read-back is intended (e.g. acknowledged 1-cycle handshake).

## Why this matters

This trap is invisible to the type system, invisible at SV emission, and only manifests in simulation as occasional spurious state transitions. The fix at source is usually a 1-line change (read the upstream port instead of the routed comb output). The cost without a lint is several iterations of debugging — including invasive thread refactors.

## Pointers

- Concrete repro / fix in [arch-ibex feedback memory](https://github.com/arch-hdl-lang/arch-ibex), pitfall #11 in `feedback_arch_syntax_pitfalls.md`.
- Related: `lower_threads` in `src/elaborate.rs` (where the dead-skid sub-state is inserted).

nogate-ai:
## Summary

Implements the dead-skid combinational-feedback lint from **issue #245** — described there as *"the most expensive single class of bug"* in the arch-ibex Phase A work (pitfall #11). A `thread` lowers to an FSM with compiler-inserted dead-skid sub-states; during those cycles the thread's **combinationally-driven** outputs fall to their default. If the thread later *reads* a signal that is a same-cycle comb function of one of those outputs — **including a path that crosses one child-instance boundary** — it can observe a stale/spurious value. The fix at source is always a 1-line change (read the upstream input, not the routed comb output); the hard part was knowing you need to. This lint makes it a compile-time warning.

The cross-module case is the one that actually hurts (the arch-ibex `IbexMultdivFast` repro and the NIC-400 thread-drives-channel-into-child pattern are all one boundary deep), and it's exactly what the intra-module-only sketch in `ideas/2026-05-28-signal-flow-graph.md` §Check 2 would have missed.

## Design

Rather than build a new signal-flow graph, this **reuses existing `comb_graph` infrastructure**:
- `comb_graph::per_output_comb_deps` / `per_output_comb_deps_fsm` already give the precise per-output → input comb dependency map (issue #246 Phase 2). One level of child-instance edges is built from these.
- `comb_graph::collect_comb_deps` (made `pub`) builds the parent comb/let adjacency.

Pipeline (`src/signal_flow.rs`):
1. `module_comb_fwd_edges` — forward comb adjacency, one boundary deep (Option B).
2. `thread_comb_write_set` / `thread_read_set` — seeds and read set. **Seeds are comb (`=`) writes only**: registered (`<=`) thread outputs *hold* across skid cycles, so a `reg x_r; comb x = x_r; wait until x;` mirror is correctly not flagged.
3. `comb_reachable_from` + `comb_path_to` — BFS reachability + shortest-path reconstruction.
4. `find_dead_skid_hazards` — ties it together, returns structured `DeadSkidHazard` records.

The lint runs on the **pre-thread-lowering AST** (so `ModuleBodyItem::Thread` is intact) in the shared check/build/sim pipeline. It emits a **warning, never an error**, naming the thread, read signal, driven signal, and comb path. Suppress an intentional read-back (e.g. an acknowledged single-cycle handshake) with the new module-level `pragma allow_dead_skid_feedback;`.

The `--emit-thread-map` HTML sidecar (PR #483) gets a **⚠ overlay**: a "dead-skid comb feedback" table next to the offending thread and an amber highlight on the read source line — turning an invisible sim bug into a visible design-review item.

## False positive found and fixed during review

Swept every bundled design. **nic400 = 0**, examples = 0, but one hit: `axi_dma_thread/ThreadMm2s` (`active`/`active_r`). Diagnosed as a false positive — `active_r` is a register mirrored by `comb active = active_r`; registered values hold across skid cycles. Root-caused to seed over-inclusion and fixed by seeding only comb-driven writes. Re-sweep: **zero false positives across all designs**; the cross-module repro still fires. A regression test pins the mirror shape.

## Scope / deferred

- **Option B, one boundary deep.** Hazards routed through ≥2 nested child levels are not reported in v1 (documented in `COMPILER_STATUS.md` and the ideas doc).
- No language-surface or emitted-SV change; pure diagnostic + visualization.

## Test plan

- [x] `cargo build` clean, no new warnings
- [x] Full suite green: **555 integration tests** + sibling suites, 0 failures
- [x] New tests: cross-module hazard, upstream-input negative, intra-module hazard, registered-mirror negative, no-threads early-out, reachability unit, pragma parse, two HTML-overlay tests
- [x] End-to-end CLI: warning fires on the repro; `pragma allow_dead_skid_feedback;` silences it; `--emit-thread-map` renders the ⚠ badge + path + highlighted line; suppressed design renders none
- [x] False-positive sweep across nic400 / axi_dma_thread / examples → 0
- [x] Rebased on latest `main`; suite re-verified green

Closes #245.

https://claude.ai/code/session_018yksZbMJ6sR2kkgWc5bKMp

---
_Generated by [Claude Code](https://claude.ai/code/session_018yksZbMJ6sR2kkgWc5bKMp)_


nogate-ai:
Note: These commits were recorded with the author identity `Claude <noreply@anthropic.com>`, which was set automatically by an AI developer tool environment setting. The code was entirely generated under my direction, and no external third-party entity holds copyright over these changes.

PR #494, opened 2026-06-03
fix(lower_threads): fold wait-until exit assignments into cond-exit arm (closes #306)

### LINKED ISSUE #306: thread: assign after `wait until` defers by 1 cycle (assigned on next state's body, not on wait-exit edge)
## Summary

`wait until cond; X <= Y;` makes `X` visible at `T+2` rather than `T+1`, where `T` is the cycle when `cond` first becomes true. The seq assign lands in the body of the next state (which becomes current at `T+1`), so its `<=` doesn't propagate until the edge at `T+2`.

This is the natural state-machine pattern but doesn't match the user-intuitive reading of "as soon as `cond` is true, set X to Y."

Surfaced D1 of arch-ibex Phase D — `FillBufferCtrl` per-FB phase tracker. The expected semantics (per AI Reference Card): the assign on the wait-exit edge should fire at exit. The actual behaviour is a 1-cycle delay, which compounded with #304 (the wait-1-cycle fix) made the FB phase pipeline 2 cycles longer than spec'd. Worked around by replacing the entire thread with a `seq + comb` block (no compiler change).

## Repro

```arch
module M
  port clk:   in Clock<SysDomain>;
  port rst_n: in Reset<Async, Low>;
  port go:    in Bool;
  port reg phase: out UInt<2> reset rst_n => 2'd0;
  thread on clk rising, rst_n low
    wait until go;
    phase <= 2'd1;        // expect: phase=1 at T+1 (T = go-detected cycle)
    wait 2 cycle;
    phase <= 2'd2;
  end thread
end module M
```

Emit:

```sv
if (_t0_state == 0) begin
  if (go) begin
    _t0_state <= 1;        // T: detect go, transition
  end
end
if (_t0_state == 1) begin
  phase <= 2'd1;            // T+1: state=1 current. phase write registered.
  _t0_cnt <= 32'(2 - 32'd1);
  _t0_state <= 2;
end
```

phase=1 visible at T+2 (1 cycle later than the natural reading).

## Expected

The assign should fold into state 0's body, gated on `go`:

```sv
if (_t0_state == 0) begin
  if (go) begin
    phase <= 2'd1;          // T: detect go AND fire assign together
    _t0_cnt <= 32'(2 - 32'd1);
    _t0_state <= 2;          // skip state 1 — its sole purpose was to hold the assign
  end
end
```

phase=1 visible at T+1, matching the user-expected timing.

## Surfaced by

arch-ibex D1 (port-ibex_icache) — `FillBufferCtrl` phase tracker. Combined with #304 (now fixed), this would have given phase advance per cycle. As a workaround the entire thread was replaced with a `seq + comb` block, dropping `thread` from D1's Phase D construct exercise. Re-introducing thread in a future swap blocked on this fix.

## Workaround

Replace `wait until cond; X <= Y;` with seq+comb gated on the same condition:

```arch
seq on clk rising
  if cond then X <= Y; end if
end seq
```

Loses the `thread` construct's expressiveness for sequential logic.

## Related

- #301 (custom-policy hook arg shadow), #302 (fsm `state` port collision), #303 (auto-inlined dep modules MODDUP), #304 (wait-1-cycle elision, fixed in PR #305) — all surfaced by D1.

nogate-ai:
## Summary

- **Bug**: `wait until cond; X <= Y;` lowered to two FSM states — the assignment fired at T+2 (one cycle too late). A skilled engineer writing hand-rolled SV would put both the condition check and the register assignment in the same `if (cond)` arm.
- **Fix**: Added a `fold_wait_until_exit_assignments` post-processing pass on `raw_states`. For any `wait until` state (S_wait) immediately followed by a sole-entry pure-action state (S_assign), S_assign's seq statements are moved into S_wait's `folded_exit_seq` field and emitted inside the `if (cond)` arm. S_assign becomes unreachable (`is_folded = true`) and is skipped during codegen.
- **Result**: The register assignment and state transition now fire on the same clock edge as cond detection, saving one cycle.

### Before / After Waveform

```sv
// Before (two states, X visible at T+2)
if (_t0_state == _t0_S0_wait_until) begin
  if (cond) _t0_state <= _t0_S1_action;
end
if (_t0_state == _t0_S1_action) begin
  X <= Y;                        // fires at T+1, visible T+2
  _t0_state <= _t0_S2_...;
end

// After (folded, X visible at T+1)
if (_t0_state == _t0_S0_wait_until) begin
  if (cond) begin
    X <= Y;                      // fires with cond detection at T, visible T+1
    _t0_state <= _t0_S2_...;    // skips S1 entirely
  end
end
```

### Safety conditions for the fold

The fold is guarded by:
1. State si must be a pure `wait until` (has `transition_cond`, no `wait_cycles`, no `multi_transitions`, no existing `folded_exit_seq` or `seq_stmts`).
2. State si+1 must be a pure action state (no `transition_cond`, no `wait_cycles`, no `multi_transitions`, not already folded), with at least one seq stmt to fold.
3. Sole-entry check: si+1 must not be targeted by any `multi_transition` (fork/join or dispatch rejoin states are excluded).
4. `no_fold_into_prev` flag: states created directly after a `wait 1 cycle` elision carry this flag, preventing the fold from losing the 1-cycle budget already consumed by the elision.

`wait N cycle` states are not eligible as si (they have `wait_cycles.is_some()`), satisfying the spec requirement that counter states remain intact.

## Test plan

- [x] All 547 pre-existing tests still pass.
- [x] `test_wait_until_fold_basic`: `wait until go; phase <= 2'd1;` — assignment appears inside `if (go)` arm, not in a separate S1 state.
- [x] `test_wait_until_fold_multi_assign`: `wait until go; X <= A; Y <= B;` — both assignments fold into the same `if (go)` arm.
- [x] `test_wait_until_fold_no_fold_past_second_wait`: `wait until go; X <= A; wait until done; Y <= B;` — X folds into S0's go-arm, Y folds into S1's done-arm; state ordering correct.
- [x] `test_wait_until_fold_wait_n_cycle_not_folded`: `wait 3 cycle; X <= A;` — counter states remain intact, X not in the go-arm.
- [x] Updated `test_thread_state_localparams_emitted` to reflect the folded 3-state shape (S1_action now unreachable, S0 jumps directly to S2_wait_cycles).
- [x] Updated `test_native_sim_thread_driven_top_pipe_reg_output_is_public` testbench: fold makes `payload_out` and `valid_out` visible 1 cycle earlier, so the check is after 1 tick instead of 2.

https://claude.ai/code/session_01NcGJqy1DDcTG66F7Fzof53

---
_Generated by [Claude Code](https://claude.ai/code/session_01NcGJqy1DDcTG66F7Fzof53)_

### DIFF
--- a/ideas/2026-06-03-auto-pipeline-asserts.md
+++ b/ideas/2026-06-03-auto-pipeline-asserts.md
@@ -1,0 +1,277 @@
+# Enhancement: `--auto-pipeline-asserts` — compiler-generated SVA hazard contracts for `pipeline` constructs
+
+**Date:** 2026-06-03
+**Status:** Proposal — needs team discussion before implementation
+**Related:** `--auto-thread-asserts` (shipped, `src/codegen/`), bounds-check SVA (shipped), div-by-zero SVA (shipped)
+
+---
+
+## Problem
+
+The `pipeline` construct auto-generates stall propagation, flush masks, and forward muxes from
+`stall when`, `flush`, and `forward` directives.  The generated logic is correct *in isolation*,
+but there is currently **no machine-checkable contract that says so**.  If a later edit to the
+pipeline body, a stall condition expression, or a downstream consumer introduces a hazard, the
+first symptom is a functional simulation failure — not a compiler or formal diagnostic.
+
+Pipeline hazard bugs are the #2 class of subtle RTL bugs after CDC, and they are particularly
+expensive to root-cause because the failure (a wrong value read N cycles after a stall or flush)
+is temporally distant from the cause (the stall/flush logic that failed to freeze/clear a stage).
+
+The ARCH compiler already auto-generates SVA contracts for three other domains:
+
+| Feature | Flag | What it asserts |
+|---|---|---|
+| Thread state transitions | `--auto-thread-asserts` | Each `wait until` / `wait N cycle` advances state correctly |
+| Vec / bit-select bounds | always-on | Index stays within declared bounds |
+| Divide-by-zero | always-on | Divisor is non-zero when expression evaluates |
+
+The `pipeline` construct is the natural next target.
+
+---
+
+## Proposed flag
+
+```
+arch build   Pipe.arch --auto-pipeline-asserts    # emit SV with embedded SVA
+arch formal  Pipe.arch --auto-pipeline-asserts    # also works (passes option to codegen)
+arch sim     Pipe.arch --auto-pipeline-asserts    # checked by Verilator via --assert
+```
+
+Off by default (same as `--auto-thread-asserts`).  All emitted SVA is wrapped in
+`// synopsys translate_off / on` so synthesis tools ignore it.  Reset polarity is inferred
+from the module's `Reset<Kind, Polarity>` port — the same helper already used by the thread
+and bounds-check SVA emitters.
+
+---
+
+## Property classes
+
+### P1 — Stall-freeze: stalled stage holds its valid and data registers
+
+When a stage is stalled (`{stage}_stall` is asserted), its registered state must not change on
+the next clock edge.  A violation means the stall backpressure logic has a bug (a hole in the
+stall chain, or a `<=` assign that bypasses the stall gate).
+
+```systemverilog
+// synopsys translate_off
+_auto_pipe_fetch_stall_freeze_0: assert property (
+    @(posedge clk) disable iff (rst)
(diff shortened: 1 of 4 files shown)


nogate-ai:
Note: These commits were recorded with the committer identity `Claude <noreply@anthropic.com>`, which was set automatically by an AI developer tool environment setting. The code was entirely generated under my direction, and no external third-party entity holds copyright over these changes.

PR #526, opened 2026-06-07
ci: keep arch-programming skill snapshots in sync with sources

nogate-ai:
The arch-programming skill bundles point-in-time copies of core ARCH docs
under skills/arch-programming/references/. Nothing kept them in sync with
their doc/ + mcp/ + root sources, and the root README snapshot had already
drifted (missing the 'AI assistant support' section).

- Add scripts/sync_skill_snapshots.sh as the single source of truth for the
  source -> snapshot mapping, with check (CI) and refresh (local) modes.
- Add the skill-snapshots GitHub Actions workflow that runs the check on PRs
  touching any source doc or bundled snapshot.
- Refresh the stale README snapshot.
- Point SNAPSHOT.md's refresh instructions at the script instead of a
  duplicated cp list.

### DIFF
--- a/.github/workflows/skill-snapshots.yml
+++ b/.github/workflows/skill-snapshots.yml
@@ -1,0 +1,49 @@
+name: skill-snapshots
+
+# The arch-programming skill bundles point-in-time copies of core ARCH docs
+# under skills/arch-programming/references/. This check fails if a source doc
+# changes without refreshing its bundled snapshot (or vice versa).
+on:
+  push:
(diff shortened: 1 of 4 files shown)


nogate-ai:
Note: These commits were recorded with the author identity `Claude <noreply@anthropic.com>`, which was set automatically by an AI developer tool environment setting. The code was entirely generated under my direction, and no external third-party entity holds copyright over these changes.

PR #539, opened 2026-06-08
feat(sim): ARCH_CXX env override for the sim C++ compiler

nogate-ai:
## Summary

`arch sim` hard-coded `g++` to compile the generated testbench/simulation binary (three `Command::new("g++")` sites in `src/main.rs`). On Linux that's real GCC, which miscompiles the C++20 coroutine testbench scheduler harc emits — so `arch sim` fails for any harc-driven testbench on a Linux/GCC system. (On macOS `g++` is a clang shim, so it happened to work locally.)

Adds a `cxx_compiler()` helper reading **`ARCH_CXX`** (default `g++`), mirroring the existing `ARCH_CXX_STD` / harc `HARC_CXX` pattern, and uses it at all three compile sites.

## Why

harc-com's new `check-backends` parity CI runs the prebuilt `arch` on Linux and hit exactly this — it currently works around it with a `g++ → clang++` PATH shim. Once a release ships this override, that shim can be dropped (`ARCH_CXX=clang++`).

This is an internal compiler-selection change — no user-facing language-surface impact.

## Verification

- Default `arch sim` (no `ARCH_CXX`) → works
- `ARCH_CXX=clang++ arch sim …` → works
- `ARCH_CXX=bogus-cxx-xyz arch sim …` → fails invoking it (`No such file or directory`), proving the override is actually wired
- Documented in `doc/COMPILER_STATUS.md` alongside `ARCH_CXX_STD`


nogate-ai:
Superseded by #538 (merged), which implements the same ARCH_CXX override. Closing as duplicate.

Issue #622, opened 2026-06-26
research: context-typed float literals (every float literal is FP32 today — awkward for BF16 and future narrow formats)


nogate-ai:
Research note from the daily review of the FP32/BF16 batch (#611–#618). This is the *general* design observation behind the point bug in #620 — filing separately so the bug can be fixed narrowly while the broader literal-typing question is tracked on its own.

## Observation

Every float literal is typed **FP32**, regardless of the surrounding context:

```arch
let f: FP32 = 1.5;          // ✓
let h: BF16 = 1.5;          // ✗  "type mismatch: expected BF16, found FP32"
let h: BF16 = (1.5).to_bf16();   // ✓  the only way to write a BF16 constant
```

So **any** BF16 constant — and BF16 is the headline LLM-inference type this feature targets — must be written as a parenthesized cast. A BF16-heavy datapath (weights, biases, clamp/threshold constants) ends up with `.to_bf16()` on essentially every literal. The integer surface has the opposite, ergonomic behavior: an untyped int literal adapts to the target width/signedness from context.

This same context-insensitivity is the root cause of the #620 codegen bug (a bare FP32 literal accepted in a BF16 reg-reset slot, then emitted as a 32-bit constant and truncated). Fixing #620 at the reg-reset site removes the silent miscompile, but the underlying "literals don't take their float type from context" gap remains everywhere else (lets, port defaults, comparisons against a BF16 signal, function-call args once those are allowed, …).

## Why it will recur

`doc/plan_fp_types.md` keeps the internal representation generic precisely so **FP16 / FP8 / MX block formats** can be added later. Each new narrow format inherits the same papercut: every constant of that type needs an explicit cast, and every literal-in-typed-slot is a potential width-mismatch surprise. Solving literal typing once, generically, pays off across all of them.

## Sketch of directions (for discussion — not committing to one)

1. **Context-typed float literals** (mirror the integer-literal rule): a float literal in a slot whose expected type is a known float type takes that type, rounding-to-nearest-even to the target format at compile time. `let h: BF16 = 1.5;` ⇒ BF16(1.5); `a_bf16 > 0.5` ⇒ compares against BF16(0.5). An ambiguous/standalone literal (no float context) stays FP32 as today. This also cleanly resolves #620 (the reset literal rounds to the reg's type). Compile-time-constant rounding is exact and deterministic.
[...]

PR #628, opened 2026-06-27
docs(fp): propagate #627 bf16-fma correction to source + plan; flag int→bf16 double-rounding

nogate-ai:
## Why

PR #627 found that the documented "bf16_fma is correctly-rounded" claim is **false** — `arch_fma_bf16` is fused **f32-accumulate** (one correctly-rounded f32 fma, then a second non-innocuous rounding f32→bf16), differing from a correctly-rounded `a·b+c` on ~0.37% of finite inputs. It corrected the two FP **READMEs** but left the **identical now-false claim** in several source comments and the design plan. That stale prose now directly contradicts a merged machine-checked theorem (`archBf16Fma_eq_narrow_roundNE`) sitting in the same repo — classic doc-drift tech debt.

## What this changes (comments + design-doc prose only — zero RTL/SMT/sim output change)

| File | Was | Now |
|---|---|---|
| `src/fp_ops.rs` | "mul/add/sub/**fma** via f32 are correctly-rounded bf16"; fma miter "spurious sat (incomplete z3 fp.fma)" | fma is **not** correctly-rounded (f32-accumulate); its `fp.fma` sat is a **genuine** counterexample. mul/add/sub claim kept (exhaustively SMT-proved). |
| `src/sim_codegen/mod.rs` | "innocuous double rounding … result equals direct correctly-rounded bf16" (sweeps fma in) | mul/add/sub correctly-rounded; **fma f32-accumulate**, not. |
| `doc/plan_fp_types.md` §5.3 + appendix | "provably single-rounded via f64", `p₁ ≥ 2p₂+2` ⇒ fma correct | correction admonition: that sufficiency argument is a **known RNE fallacy** (fails at `p₁=4,p₂=1`: `RNE₁(RNE₄(1.4375))=2` ≠ `RNE₁(1.4375)=1`); shipped uses an **f32** intermediate, not f64; f32 mul/fma now **Lean-proved** (#625/#626). |

## New finding (separate bug, same class) — flagged, not fixed

While auditing, I found **`int.to_bf16()` has the same double-rounding defect**: `int → f32 → bf16` is **not** correctly-rounded for `|i| ≥ 2²⁴`.

- Witness: `i = 16842753` (`= 2²⁴+2¹⁶+1`) → `0x4b80`, correctly-rounded is `0x4b81`. `(float)i` ties-to-even down onto the exact bf16 midpoint, then the narrow ties the wrong way.
- **8064** mismatching integers in `[1, 2³⁰)`, all ≥ 2²⁴, each off by 1 bf16 ULP.
- Both backends agree (SV `arch_f32_to_bf16(arch_*_to_f32(...))` and sim `_arch_{i,u}_to_bf16`), so it's not a sim/SV divergence — it's a shared semantic question.

Routing via f32 is hardware-realistic (RISC-V has no direct int→bf16), so this **may** be intended-as-is — but the result is not correctly-rounded and that should be a deliberate, documented decision, not an accident left under an "innocuous" comment. **Behavior is unchanged here** (the conversion result is user-facing semantics); only the comment is corrected to be honest. Filed as a separate issue for a spec decision.

## Verification

`cargo build --release` clean. No `.sv` / `.smt2` / sim-C output bytes change — confirmed the edits are restricted to `//` comments and `.md` prose.

### DIFF
--- a/doc/plan_fp_types.md
+++ b/doc/plan_fp_types.md
@@ -170,6 +170,23 @@ saturate per the RISC-V spec (§6).
 
 ### 5.3 BF16 operations — provably single-rounded via f64
 
+> **⚠ Correction (PR #627) — this section is superseded and its `fma` claim is
+> unsound.** Two things diverge from what shipped:
+> 1. **Intermediate precision.** The implementation does **not** use an f64
+>    intermediate (`p₁ = 53`); it uses an **f32** intermediate (`p₁ = 24`,
+>    `widen → arch_*_f32 → narrow`, see §appendix and `src/fp_ops.rs`).
+> 2. **`fma` is not innocuous.** The `p₁ ≥ 2·p₂ + 2` sufficiency argument below
+>    is a **known fallacy** for round-to-nearest (it fails already at
+>    `p₁ = 4, p₂ = 1`: `RNE₁(RNE₄(1.4375)) = 2` but `RNE₁(1.4375) = 1`). It does
+>    hold operationally for bf16 `mul`/`add`/`sub` — those are *exhaustively*
+>    SMT-proved correctly-rounded vs `fp.{mul,add,sub}` on `(8,8)` — but **not**
+>    for `fma`. The shipped `arch_fma_bf16` is fused f32-accumulate (one f32 fma,
+>    then a second rounding to bf16) and differs from a correctly-rounded
+>    `a*b+c` on ~0.37% of finite inputs, always by 1 ULP. See §8 below,
+>    `tests/fp_v1/smt_proof/README.md`, and `proofs/lean_fp_equiv`.
+>
+> The rest of this section is retained as historical design rationale.
+
 SoftFloat has no native bfloat16. Naively doing a BF16 op by promoting to
 `float32`, operating, and narrowing **double-rounds** and can be wrong. We avoid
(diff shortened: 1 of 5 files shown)


nogate-ai:
The int→bf16 finding is now tracked in #629 (with the witness, scope, and the document-vs-fix decision). The corrected `src/codegen/mod.rs` comment cross-references it.

nogate-ai:
Added `doc/proposal_fp_rounding_semantics.md` — a research note generalizing this PR's two findings (#627 bf16_fma, #629 int→bf16) into a recommendation: the spec should normatively classify every FP op/conversion as **correctly-rounded (CR)** or **convention-rounded-via-f32 (VR(f32))**, so 'correctly-rounded' becomes a deliberate, verification-load-bearing claim instead of an unexamined default the `p₁≥2p₂+2` fallacy keeps re-introducing for each new narrow format (fp16/fp8 on the horizon, #622). Note only; no implementation.

FP test suite (`cargo test --release fp`) + integration tests: **17 passed, 0 failed** — including `fp_smt_equivalence_proofs` and `fp_rtl_differential_equiv_verilator`, which validate the bf16 mul/add/sub correctly-rounded claims this PR *keeps*.

PR #643, opened 2026-06-30
fix(#636): param-width sized literal — type as UInt<W> + emit valid SV cast

nogate-ai:
Two bugs in the param-width sized-literal feature shipped in #636 (`W'd0`, `ADDR_WIDTH'd42`, …). Found during the 2026-06-29 daily PR review. The feature was **non-functional for its headline use case** and **emitted invalid SystemVerilog**.

---

## Bug 1 — type checker types `W'dN` as `UInt<32>`, not `UInt<W>`

The PR's own motivating example fails type-check for any `W != 32`:

```arch
reg r: UInt<W> reset rst => W'd0;
r <= W'd0;   // × width mismatch: `r` is UInt<8> but RHS is UInt<32>
```

**Root cause.** `elaborate` runs before `typecheck`, but `subst_expr_params` (which resolves `ParamSized → Sized`) only fires when a module is monomorphized as a sub-instance. A top-level module's `param W` default is never substituted into expression literals, so `W'd0` reaches the checker as `LitKind::ParamSized`, and the arm at `typecheck.rs:3644` hardcoded `Ty::UInt(32)`, ignoring the width — even though the *type* `UInt<W>` is resolved correctly. The asymmetry is the bug.

**Fix.** Resolve the width identifier via the existing `eval_const_expr` helper (same mechanism that resolves `UInt<W>`). `W'dN` now types as `UInt<W>`, behaving identically to a concrete `W'dN`: lossless widening into a wider target allowed (= `8'd5`→`UInt<16>`), narrowing still needs explicit `.trunc<N>()` (= `32'd5`→`UInt<16>` error). Falls back to `UInt<32>` only when the param is genuinely unbound.

| case | before | after |
|---|---|---|
| `W'd0` → `UInt<W>`, W=8 | ❌ error | ✅ ok |
| `W'd5` → `UInt<W>`, W=48 | ✅ ok | ✅ ok |
| `W'd5` (W=8) → `UInt<16>` | ❌ error | ✅ ok (lossless, = `8'd5`) |
| `W'd5` (W=32) → `UInt<16>` | ❌ error | ❌ error (needs `.trunc`) ✓ |

---

## Bug 2 — codegen emits invalid SystemVerilog `W'd5`

A param-width literal that survives to SV codegen (top-level module keeps `param W` symbolic) was emitted as `W'd5`:

```sv
assign o = W'd5;
// %Error: syntax error, unexpected INTEGER NUMBER  (Verilator 5.048)
```

This is **not legal SystemVerilog** — a sized-literal width must be a constant *number*, not a parameter. Pre-existing in #636 and independent of Bug 1 (the `W=32` case emitted invalid `W'd5` on `main` too).

**Fix.** Emit the legal parameterized form — a size cast `W'(5)` — which carries the same value. Only `src/codegen/mod.rs` changes; `interface.rs` (`.archi` files are *ARCH* syntax, where `W'd5` is correct) and `thread_map.rs` (debug labels) are deliberately untouched.

Verified: `verilator --lint-only` clean; `arch sim` yields `o=5` for `W=8`.

---

## Tests / docs

- `tests/integration_test.rs`: **+3** regression tests (type as `UInt<W>`; over-width still rejected; SV emits `W'(5)` not `W'd5`).
- Full `cargo test --release` green except the two pre-existing Lean/elan env failures (`construct_proof_lean_*` — no toolchain locally).
- `cargo fmt --check` clean.
- #636 shipped the feature undocumented; this adds param-width-literal + `local param`-in-`ram`/`counter`/`arbiter` notes to the spec + reference card and refreshes the skill snapshots.

---

## ⚠️ Maintainer sign-off requested (Bug 1 only)

Per `CLAUDE.md`, **Bug 1** touches type-system surface behavior (the type inferred for a literal). It's bug-completion — `W'dN` becomes consistent with the concrete-`N'dN` rule and the feature's documented intent — but flagging for confirmation. **Bug 2** is a pure internal codegen fix (emit valid SV), autonomously in scope.

### DIFF
--- a/doc/ARCH_HDL_Specification.md
+++ b/doc/ARCH_HDL_Specification.md
@@ -722,6 +722,10 @@ A module is the fundamental unit of design in Arch. Every module follows the sam
 | //                                  `NAME[i]` reads as a packed part-select.   |
 |                                                                                |
 | // local param NAME: const = expr;  // derived — emits `localparam` (not overridable) |
+| //                                  // valid in module, ram, counter, arbiter |
+| //                                                                             |
+| // A param-width sized literal `NAME'dV` (e.g. `W'd0`, `ADDR_WIDTH'd42`) is    |
+| // typed `UInt<NAME>` — same widening rules as a concrete `8'd0`.              |
 |                                                                                |
 | //                                                                             |
 |                                                                                |
--- a/doc/Arch_AI_Reference_Card.md
+++ b/doc/Arch_AI_Reference_Card.md
@@ -331,6 +331,7 @@ Call:        FnName(arg1, arg2)    // overload-resolved by argument types
 Enum:        E::Variant
 Struct lit:  S { f: val }
 Sized lit:   8'hFF  16'd1024  4'b1010   // Verilog-style
+Param lit:   W'd0  ADDR_WIDTH'd42        // param-width prefix → typed UInt<W>
 todo!        // compilable placeholder; warns at compile, aborts at sim runtime
 ```
 
--- a/skills/arch-programming/references/ARCH_HDL_Specification.md
+++ b/skills/arch-programming/references/ARCH_HDL_Specification.md
@@ -722,6 +722,10 @@ A module is the fundamental unit of design in Arch. Every module follows the sam
 | //                                  `NAME[i]` reads as a packed part-select.   |
 |                                                                                |
 | // local param NAME: const = expr;  // derived — emits `localparam` (not overridable) |
+| //                                  // valid in module, ram, counter, arbiter |
+| //                                                                             |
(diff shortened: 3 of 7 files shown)


nogate-ai:
Closing as superseded and conflicting. The two bugs this PR bundled have each landed via a narrower, non-conflicting change:

- **Bug 1** (typecheck: `W'dN` typed as `UInt<32>` instead of `UInt<W>`) — fixed by #651 (merged).
- **Bug 2** (codegen: invalid SV `W'd5`) — fixed by #652 (open, CLEAN/mergeable) via the size-cast form `W'(5)`.

This PR is now DIRTY/CONFLICTING against main and carries no unique content. Closing to keep the backlog clean; #652 remains the live carrier for Bug 2.

Issue #653, opened 2026-07-07
Variable part-select `(expr)[s +: w]` silently miscompiles; #651 slice-portability guardrail + codegen paren-list have drifted


nogate-ai:
Found during the 2026-07-06 automated daily review, examining #651 ("Prepare ARCH 0.70.6 Icarus portability fixes"). #651 added `is_portable_bit_slice_base` in `typecheck.rs`, which rejects `arch check` of `(expr)[hi:lo]` bit-slices on non-portable bases. Investigating that guardrail surfaced three related issues — one is a **silent miscompile**.

## 1. (Severe) Variable part-select on an expression base silently miscompiles

The #651 guardrail only guards `ExprKind::BitSlice` (`[hi:lo]`). The sibling **variable part-select** `ExprKind::PartSelect` (`[start +: w]` / `[start -: w]`) is **not** guarded at typecheck, **and** its codegen (`src/codegen/mod.rs` ~line 6213) never parenthesizes its base. So a low-precedence base produces precedence-wrong SV with no diagnostic:

```
module PsTest
  port a: in UInt<8>; port b: in UInt<8>; port s: in UInt<3>; port y: out UInt<4>;
  let y = (a + b)[s +: 4];
end module PsTest
```
```sv
assign y = a + b[s +: 4];   // parses as a + (b[s +: 4]) — WRONG
```
```
module ShPs
  ... let y = (a >> 1)[s +: 2];
```
```sv
assign y = a >> 1[s +: 2];  // parses as a >> (1[s +: 2]) — WRONG
```

`arch check` accepts both; `arch build` emits silently-wrong SV. This is strictly worse than the BitSlice case, which #651 at least rejects. Note the correct parenthesized form is **not** an escape either — Verilator rejects `(a + b)[s +: 4]` as a syntax error — so the only valid lowering is a named intermediate (exactly what #651's message tells users to do for BitSlice).

**Fix options (maintainer decision — user-facing):**
- **(A, consistent with #651)** extend the `is_portable_bit_slice_base` rejection to the `PartSelect` arm in `typecheck.rs`, reusing the same diagnostic. Turns the silent miscompile into the same clear error. Requires a spec/MCP-guidance line like the BitSlice one.
- **(B)** codegen auto-hoists the base into a temp wire (`logic [..] __t = a + b; y = __t[s +: w];`). Internal-only, keeps the construct working, but diverges from #651's "reject + name it" convention.

[...]

PR #660, opened 2026-07-11
fix(typecheck): guard variable part-select base against silent miscompile (arch#653 item 1)

nogate-ai:
## Summary

Closes item 1 of #653 — a **silent miscompile**. A variable part-select `(expr)[start +: w]` on a low-precedence base passed `arch check` and emitted precedence-wrong SV with no diagnostic:

```arch
let ps = (a + b)[2 +: 3];
```
```sv
assign ps = a + b[2 +: 3];   // parses as a + (b[2 +: 3]) — the select binds to b alone. WRONG.
```
`(a >> 1)[s +: 2]` similarly emitted `a >> 1[s +: 2]`.

## Root cause

#651 added `is_portable_bit_slice_base` and wired it into the `ExprKind::BitSlice` arm of `resolve_expr_type`, so `(a + b)[hi:lo]` is correctly rejected. But the sibling `ExprKind::PartSelect` arm bound `_base` and never consulted the guard — so part-selects of the same non-portable bases slipped through and miscompiled. The parenthesized form is not an escape either (Verilator/iverilog reject `(a + b)[start +: w]`), so the only valid lowering is a named intermediate.

## Fix

Mirror the BitSlice guard onto the PartSelect arm, reusing `is_portable_bit_slice_base` **unchanged**, with a part-select-specific message:

```
cannot part-select this expression directly; SystemVerilog backends cannot
portably emit `(expr)[start +: width]`. For same-width modular arithmetic,
use wrapping operators such as `+%` or `-%`; otherwise assign the
expression to a typed `let`/wire first and part-select the named value.
```

Portable bases are unaffected — `sum[s +: 3]` (named), `a[i][s +: w]` (Index base), `{a,b}[s +: w]` (Concat), etc. still type-check and emit the bare `name[start +: w]`.

## Scope & blast radius

- **User-facing** (new typecheck rejection) — implemented with maintainer sign-off, mirroring the already-approved #651 BitSlice convention. `is_portable_bit_slice_base` itself is untouched.
- **Item 2 of #653** (reconciling the portable set with codegen's no-paren list for *chained bit-slice* bases like `a[7:4][1:0]`) is a separate, lower-severity portability question — **left open**, not in this PR.
- **Zero blast radius:** no `.arch` fixture in `tests/` or the `cvdp`/`realbench`/`backend_equiv` corpora part-selects a non-portable base (verified by grep).

## Verification

- `cargo build --release`
- `cargo test --release --test integration_test -- part_select bit_slice_arithmetic` → 4/4 pass
  - `test_part_select_arithmetic_expression_errors_with_wrapping_hint` (arith + shift bases rejected)
  - `test_part_select_named_value_base_accepted` (named base still emits bare `sum[s +: 3]`)
  - existing `test_bit_slice_arithmetic_expression_errors_with_wrapping_hint` still green
- End-to-end: `(a+b)[2+:3]` now errors at `arch check`; `sum[s +: 3]` builds + is Verilator-clean.
- Docs: `doc/COMPILER_STATUS.md` + skills snapshot updated (part-select guard).
- Full regression sweep running post-open per commit-before-verify ordering; will follow up if anything surfaces.

Found + fixed during the 2026-07-10 automated daily review.

### DIFF
--- a/doc/COMPILER_STATUS.md
+++ b/doc/COMPILER_STATUS.md
@@ -5,7 +5,7 @@
 >
 > **0.70.6 release highlights:**
 > - **Icarus-portable CVDP generation guidance** — ARCH MCP and skill instructions now steer generated code away from indexed casts/conversions, nested signed-extension chains, and direct slices of arithmetic expressions that Icarus rejects or mishandles.
-> - **Safer type checking for portable slices** — `arch check` now rejects direct bit-slicing of non-portable expression bases such as arithmetic expressions and points users toward wrapping arithmetic or named typed intermediates.
+> - **Safer type checking for portable slices** — `arch check` now rejects direct bit-slicing *and variable part-selecting* (`(expr)[start +: w]`) of non-portable expression bases such as arithmetic expressions and points users toward wrapping arithmetic or named typed intermediates. (The part-select guard closes a silent miscompile where `(a + b)[start +: w]` emitted `a + b[start +: w]`.)
 > - **Parameterized sized literals** — literals such as `W'd0` now resolve to the positive integer value of `W` instead of falling back to `UInt<32>`, with a clear diagnostic for non-positive widths. SV codegen emits the legal size-cast form `W'(0)` (a parameter is not a valid sized-literal width — Verilator and iverilog both reject `W'd0`).
 >
 > **0.70.4 release highlights:**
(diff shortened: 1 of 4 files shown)


nogate-ai:
Full regression sweep complete on a fresh worktree build: **694/694** integration tests + **50/50** lib tests pass, 0 failures. No follow-up needed.

nogate-ai:
Superseded by #659, which merged to main first and closes #653 more completely — it covers **both** item 1 (the PartSelect guard, near-identical to this PR) **and** item 2 (removing `BitSlice`/`PartSelect`/`Bool`/`EnumVariant` from `is_portable_bit_slice_base`), plus the spec/reference-card/MCP doc updates.

Verified #659 on a fresh main build: `(a+b)[2+:3]` and `a[7:4][1:0]` are both now rejected with clear messages; named/portable bases still emit bare; its slice tests pass. I also confirmed the item-2 design is correct — iverilog rejects bare `a[7:4][1:0]` ("All but the final index in a chain of indices must be a single value, not a range"), so rejecting at typecheck (rather than emit-bare, which is Verilator-only) is the portability-correct call. Closing this as a duplicate.

Issue #501, opened 2026-06-04
Thread surface gaps: shipped vs. missing features inventory


nogate-ai:
Tracking issue collecting the gaps in the current ARCH thread surface, drawn from `doc/thread_spec_section.md`, `doc/thread_multi_outstanding_spec.md`, `doc/thread_pipeline_spec.md`, `doc/plan_thread_parallel_sim.md`, and `doc/plan_thread_parallel_sim_phase3.md`. Some items are intentionally deferred (HLS path, FSM-construct shared work); they're listed so triage starts from one place rather than three docs.

## What's shipped (no action needed — for orientation)

- Core syntax: `thread` / `thread once`, named threads, `wait until`, `wait N cycle`, `wait 0+ cycle until` (Mealy), `do..until`, `fork`/`join`, `for` loops with `wait`, `if/elsif/else` with `wait` in either branch (v0.45.0+), `generate for/if` over threads.
- Trailing-seq merge into preceding `wait/do..until` exit (no dead cycle); `do..until` inside `lock` fires same cycle the grant lands.
- `resource R: mutex<policy>` with all `arbiter` policies (round_robin, priority, lru, weighted, Custom via `hook grant_select`) — synthesized as a real `ArbiterDecl` inside the merged threads module.
- `shared(or)` for both comb-driven and seq-driven signals.
- Pipeline integration (v0.40.0): `wait until` / `do..until` inside pipeline-stage `seq` blocks wired into stall chain + flush.
- TLM integration (§20.11): target-method threads, serialized initiator calls in counted `for` and `if/elsif/else`, RHS-fork issue groups with compute-only tail.
- `--auto-thread-asserts` SVA (§20.15) with proof corollaries in `doc/thread_lowering_proof.md`.
- Pre-lowering coroutine sim (`--thread-sim parallel`, Phases 1+2: #143–#152), bit-identical to fsm path on cross-checked `tests/thread/` cases.

## Language-feature gaps

- **`semaphore<N, policy>` (§20.8.4) — declared planned, not parsed.** `parse_resource_decl` (`src/parser.rs:2433`) accepts only `mutex<…>`. No carrier in `elaborate.rs` either. Use case: limit concurrency (e.g. only 4 of 32 threads in AR phase) without strict mutex exclusion.
[...]

PR #687, opened 2026-07-13
feat(threads): re-enable comb overlap optimization with lock-state tagging (part of #501)

nogate-ai:
## Summary

Implements the codegen-TODO item from #501: re-enables the comb overlap optimization for lowered threads, disabled since the zero-cycle-lock change (c4b77f0) because it could leak lock-guarded outputs into preceding states.

- `lower_thread_lock` now tags every state it produces with a new `ThreadFsmState::is_lock_body` flag (including the empty-body fallback state).
- The overlap optimization is re-enabled at the old TODO site in `lower_module_threads`: for a state with a single conditional transition, the next state's comb outputs are additionally driven under `state == si && trans_cond` during the transition cycle, removing the dead cycle between back-to-back thread states (e.g. back-to-back handshakes: `r_ready` now asserts during the `ar_ready` cycle in the AXI read-loop fixture).

## Safety guards
- **Next state is a lock body → no overlap.** Grant-gated outputs never appear before the arbiter grant cycle. This is the #501 fix: lock states are now tagged, so the optimization is provably scoped away from critical sections.
- Multi-transition (fork/join, if-dispatch) and TLM terminal-return states are skipped — their successor is condition-dependent.
- Folded states (#306) are skipped; `folded_exit_target` is used as the effective successor, mirroring the seq-side next-state computation exactly.
- Terminal once-state self-loops are skipped.

## Testing
- Full `cargo test --release` green (the only failures in my run were 2 Lean-replay tests failing for environment reasons — fresh worktree without `lake build`; re-verified separately).
- New regression tests:
  - `test_thread_comb_overlap_drives_next_state_outputs` — overlap arms present on the AXI read-loop fixture.
  - `test_thread_comb_overlap_never_enters_lock_body` — a lock-heavy two-writer mutex fixture emits **zero** overlap arms (every transition lands in a lock body).
- Long verification (sim⇄Verilator lock-step on lock-heavy fixtures + thread TB sims) running post-push; follow-up commit if anything surfaces.

Part of #501 (one item of the tracking issue — do not auto-close).

### DIFF
--- a/doc/thread_spec_section.md
+++ b/doc/thread_spec_section.md
@@ -504,8 +504,9 @@ The overlap is Mealy-style: the next state's outputs respond to the transition c
 
 The overlap also does not apply to:
 - multi-transition states (fork/join dispatch, `if`/`else` dispatch) — the successor is condition-dependent
-- TLM target return states
 - terminal states of `thread once` (self-loop; nothing to overlap)
+
+TLM method target threads are lowered by a dedicated response-router path and do not participate in the overlap.  As with each state's own comb arm, a thread's `default_when` condition does not gate the overlap arm — `default_when` preempts the *seq* side (state register and registered assigns) only.
 
 ## 20.16  Auto-Emitted Spec-Contract SVA (`--auto-thread-asserts`)
 
--- a/doc/thread_spec_section.md
+++ b/doc/thread_spec_section.md
@@ -475,7 +475,39 @@ The merge fires `ar_done_r <= 1` on the same clock edge as the `ar_ready` handsh
 
 If the seq assign appears after a multi-transition state (e.g. inside a `for` loop), the compiler uses the loop exit condition as the guard instead of every iteration.
 
-## 20.15  Auto-Emitted Spec-Contract SVA (`--auto-thread-asserts`)
+## 20.15  Comb Overlap Across State Transitions
+
+When a state's single conditional transition fires (`wait until cond` or `do..until cond`), the compiler additionally drives the **next state's comb outputs during the transition cycle**.  This removes the dead cycle between back-to-back states:
+
+```arch
+thread on clk rising, rst_n low
+  ar_valid = 1;
+  ar_addr  = 32'd100;
+  wait until ar_ready;   // ← the cycle ar_ready is high, r_ready is ALREADY driven
+
+  r_ready = 1;
+  wait until r_valid;
+  data_r <= r_data;
+end thread
+```
+
+The generated comb block contains an overlap arm alongside the normal per-state arm:
+
+```sv
+if (_t0_state == _t0_S1_wait_until) r_ready = 1;                 // normal
+if (_t0_state == _t0_S0_wait_until && ar_ready) r_ready = 1;     // overlap
+```
+
+The overlap is Mealy-style: the next state's outputs respond to the transition condition combinationally, one cycle before the state register advances.  Register values read by those outputs are the *pre-transition* values (the transition edge has not fired yet).
+
+**The overlap never applies when the next state is inside a `lock` body.** Lock-body outputs are gated by the arbiter grant (§20.8, §20.13); driving them from the preceding state would leak critical-section signals before the grant cycle.  Transitions *into* a lock block therefore always take the full cycle — the lock's own zero-cycle grant path (§20.13) is the mechanism that removes lock-entry latency, not the overlap.
+
+The overlap also does not apply to:
+- multi-transition states (fork/join dispatch, `if`/`else` dispatch) — the successor is condition-dependent
+- TLM target return states
+- terminal states of `thread once` (self-loop; nothing to overlap)
+
+## 20.16  Auto-Emitted Spec-Contract SVA (`--auto-thread-asserts`)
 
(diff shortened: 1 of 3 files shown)


nogate-ai:
Long verification complete (all on the freshly built branch binary):

1. **Full `cargo test --release`** — green. The 2 formal_test failures in the first run were environmental (fresh worktree without `lake build`); after building the Lean library, formal_test is 23/23.
2. **Nightly equivalence sweep** (`scripts/nightly_equivalence.py`) — 177/177 units pass on first pass (check → build → Verilator lint → arch sim with real TBs where manifested, including `thread__basic_thread`, which now exercises the overlap arms), plus 16/16 backend_equiv torture fixtures.
3. **sim⇄Verilator lock-step** (`harc sim --check-backends`, same seed, per-cycle semantic trace diff) on the three lock-heavy fixtures:
   - `Fx4MutexContendRR` — PASS, total=30 g0=10 g1=10 g2=10, **traces match, no divergence**
   - `Fx10MutexContendPrio` — PASS, total=30 g0=30, **traces match, no divergence**
   - `Fx2GenThreadLock` — PASS, grants=31, **traces match, no divergence**

Also added spec §20.15 documenting the overlap timing semantics and the lock-body exclusion (doc commit 13ca770).

nogate-ai:
Independent model review (Sonnet agent, read-only, fresh binary + 5 probe fixtures): **approve-with-nits, no correctness/soundness bugs found**. Edge cases explicitly verified: lock-followed-by-lock emits no overlap arm; fork/join dispatch and branch-redirected states are excluded via the `multi_transitions` guard (leveraging the pre-existing redirect invariant); `thread once` terminal self-loop emits nothing; #306 folding never resurrects a folded state's comb; `wait_cycles` states never carry a `transition_cond` (shape doesn't exist); overlap-vs-own-arm ordering is sound (mutually exclusive state guards within a thread; shared(or) OR-accumulation is order-insensitive); `lower_thread_lock` is the only producer of grant-gated states, so the tagging is complete.

Two doc/clarity nits, both addressed in 9a07e60:
- `terminal_return` guard is defensive (TLM target threads never reach this loop today) — comment updated, misleading doc bullet replaced.
- §20.15 now notes that `default_when` does not gate comb arms (matching the pre-existing per-state behavior) and that TLM targets use a separate lowering path.

nogate-ai:
Daily review sweep (2026-07-15): flagged `awaiting-decision` — this PR is stale (25 commits behind `main`, `mergeStateStatus: DIRTY`) and its rebase raises a **semantic**, not textual, conflict that needs a maintainer decision rather than an autonomous reconcile-and-merge.

**The conflict.** `git rebase origin/main` collides in `src/elaborate.rs` (4 hunks). The core one is the lock-state metadata struct: this PR inserts `is_lock_body: bool`, but #691 (merged since this branch forked) inserted `lock_release: Option<String>` at the same location. The two fields are not mutually exclusive — the struct can carry both — but resolving it correctly means more than keeping both lines:

- **#687 predates #691** (hold-stable lock arbiter + same-cycle lock timing). The `is_lock_body` gating this PR adds to suppress comb-overlap on lock-body states must be reconciled with #691's `lock_release` pulse mechanism (back-to-back re-lock detection). They touch the same grant/hold latch behavior.
- **#687 predates #701** (eval() delegates comb settling; `settle_depth` model changed from the old O(d²) loop to explicit 3/4-pass settling). The comb-overlap optimization this PR *re-enables* interacts directly with the settle model — #701 showed the settle-depth analysis was previously undercounting, and comb-overlap changes which comb blocks run in which pass. The optimization's correctness needs re-validation against the new settle model, not just a textual rebase.

**Recommendation.** Rebase onto current `main` keeping both struct fields, then re-verify the comb-overlap re-enable against the post-#691/#701 lock + settle behavior (the mutex/semaphore fixtures + `--thread-sim both` cross-check + a `--debug` trace diff vs baseline are the relevant gates). Because this is correctness-critical thread-lowering codegen and a feature (not a live-bug fix), the daily sweep is not auto-merging it. It remains part of #501.

nogate-ai:
Rebased onto current main (3ef0c72) and **fully re-validated against the post-#691/#701 semantics** (maintainer approved proceeding). This was a semantic rebase, not a blind one:

- `is_lock_body` now sits alongside #691's `lock_release`/`lock_release_info` in `ThreadFsmState`; tagging composes cleanly with the release-pulse mechanism (all conflicts were both-sides-added-fields; one rename `shared_or_signals` → `shared_signals`).
- Probed the new interaction directly: overlap from a lock-**release** state into a following non-lock state uses the full transition condition **including grant**, and the registered `_<res>_release_<ti>` pulse fires on the same edge untouched. Wrap-around into a lock's first body state still emits no overlap arm (no early `req`/output leak).
- The overlap arm reads the arbiter's comb `grant` — a cross-module comb dependency now handled by #701's self-settling `eval_comb`; confirmed empirically below.

Re-validation results on the rebased branch:
- Full `cargo test --release`: green (745 integration tests + all sub-suites) — including main's new #691/#714 lock-timing tests and the #696 semaphore ≤N churn bound, all passing **with the overlap enabled**.
- Nightly equivalence sweep: **179/179** units, **16/16** backend_equiv torture fixtures.
- `harc sim --check-backends` lock-step (same-seed per-cycle trace diff) on Fx4MutexContendRR / Fx10MutexContendPrio / Fx2GenThreadLock: **no divergence** under the new hold-stable-grant + release-pulse arbiter.

Removing `awaiting-decision`; PR is mergeable again.

PR #691, opened 2026-07-13
fix(threads): hold-stable lock arbiter grants + faithful same-cycle lock timing in both sim backends

nogate-ai:
## Summary

Investigating the `--thread-sim both` cross-check failure for `mutex<policy>` `lock` blocks whose body ends with `wait until <release>` (reported from the semaphore-resource PR investigation) turned up **three distinct bugs** — including an SV-level mutual-exclusion miscompile that is more serious than the reported timing skew.

### Bug 1 — SV miscompile: lock grants were not hold-stable (mutual exclusion + liveness)

The synthesized lock arbiter (`_arb_<Mod>_<res>`) reused the standalone `arbiter` construct's per-cycle transactional grant semantics, but a lock requester holds `request_valid` for the whole multi-cycle critical section:

- **round_robin**: `rr_ptr_r` advances every granted cycle, so with two contenders the grant **ping-pongs between them mid-hold** — both threads' lock-body comb executed in alternating cycles, and a 1-cycle release pulse was missed whenever it landed on a cycle the holder didn't own the grant. Reproduced under **Verilator 5.048 on the emitted SV** (not just arch sim): `busy1` asserted while T0 was still inside its critical section.
- **priority**: a later-arriving lower-indexed requester preempted a live holder the same way.

Both violate the "once granted … cannot lose the grant" step of Lemma L (`doc/thread_lowering_proof.md` §II.7) — that proof step was **unsound as implemented**.

**Fix**: `ArbiterDecl::lock_hold` (set only by `synthesize_lock_arbiter`; user arbiters unchanged) adds an owner latch (`hold_valid_r`/`hold_owner_r`) ahead of every policy scan in both the SV emitter and the sim arbiter model. The owner keeps the grant while its request stays asserted; re-arbitration is combinational the cycle the request deasserts, so **release → re-grant remains same-cycle**.

### Bug 2 — arch-sim FSM path lagged its own SV by one cycle

A child module's `eval_comb()` ran its internal inst-feedback loop (thread req → arbiter inst → grant → thread comb) only **one pass** per call; the internal settle loop existed only in `eval()`, which is never invoked for child instances. Grants lagged requests by one pass → the reported 1-cycle release→re-grant skew (and 1-cycle-late acquisitions) **relative to Verilator on the same SV**. `eval_comb()` now settles internally to the module's own `settle_depth`, making the FSM sim path cycle-exact with Verilator on all repro designs.

### Bug 3 — coroutine path: 1-cycle-late acquisition + order-dependent handoff

`--thread-sim parallel` burned one scheduler tick on every lock acquisition even when the resource was free (the FSM fuses request+grant+body-comb into the arrival cycle), and same-tick release→re-grant only propagated from lower- to higher-indexed threads (sequential per-thread scheduler order). New `lock_wait_until` awaiter permits **one same-tick re-fire per slot** (capped, matching the FSM's combinational grant), and the sequential tick iterates the scheduler group to a fixed point (`tick_begin()` / `tick_run()` split) so handoff is order-independent. Regular `wait until` keeps the min-1-cycle rule. mt mode (`--threads N>1`) unchanged, documented as non-timing-authoritative.

## Answer to the reported questions

- The round_robin "oscillation without completion" was **not** the same 1-cycle skew — it was Bug 1, a genuine SV-level mutual-exclusion/liveness miscompile, now fixed.
- The 1-cycle skew was **not inherent** to FSM state-register timing (the hypothesis in the report): by SV semantics the releasing thread's req drops combinationally in its exit state, so re-grant was always same-cycle in the SV — the skew was Bug 2, a sim infidelity. No permanent cross-backend timing difference needs documenting; the backends now agree cycle-for-cycle and the docs state the contract.

## Verification

- 4 new fixtures × `--thread-sim both` cross-check PASS: release-regrant (rr + priority) and hold-under-preemption/reverse-handoff (rr + priority); the preempt TBs hard-fail on any mid-hold grant migration.
- Verilator 5.048 lock-step: emitted SV matches the arch-sim FSM trace cycle-for-cycle on all fixtures.
- Golden test: `hold_valid_r` present in synthesized lock arbiters, absent from user-declared `arbiter` constructs.
- Full `cargo test --release`: green except 2 pre-existing `formal_test` Lean-toolchain environment failures that also fail on unmodified `f028f02` (verified via stash).

## Docs updated (proof-load-bearing)

- `doc/thread_lowering_proof.md` — Lemma L argument rewritten around the hold latch (previous argument was unsound for mid-hold preemption).
- `doc/thread_lowering_algorithm.md` — hold-latch description; liveness-per-policy and mutual-exclusion sections corrected; nested-lock ban rationale now rests on hold-and-wait deadlock (its old motivating scenario is closed by the latch).
- `doc/thread_spec_section.md` §20.8.3, `doc/ARCH_HDL_Specification.md` §7a.3 policy table, `doc/plan_thread_parallel_sim.md` (new lock timing contract section).

Internal codegen/lowering/sim fix — no user-facing syntax or semantics change beyond making the implementation match the already-documented exclusive-hold semantics.


nogate-ai:
**Follow-up commit `1fb7c43`** — the first CI run caught a real regression the hold latch introduced: `test_axi_read_beat_interleave_{arch_sim,verilator}_behavior` failed because a hold latch keyed on `request_valid` alone cannot distinguish *still holding* from *released and immediately re-requesting*. A thread that re-locks back-to-back in a loop (the documented beat-interleaving idiom in `ThreadAxiReadBeatInterleave.arch`) never deasserts its request wire across the `end lock` boundary, so the grant never rotated to the waiting lane.

Fix: each lock's **last body state now pulses a per-thread `request_release[i]`** combinationally when its exit transition fires (new `ThreadFsmState::lock_release` marker; relocates into the absorbing predecessor's cond-exit arm under the issue-#306 fold; covers transition_cond / wait_cycles / multi_transitions / unconditional exits). The arbiter's hold ff clears on `release[grant_requester]`, so the boundary re-arbitrates from `rr_ptr = owner+1`: contenders rotate per beat, and a lone requester re-wins the free scan with **no bubble**. Mid-body hold-stability and the same-cycle req-drop re-grant path are unchanged; user arbiters get no release port.

Injected release stmts reuse spans from the state's existing exit condition so thread-map source bands are unperturbed.

**Verification after follow-up:** full workspace suite green — 727/727 integration (both beat-interleave variants, all four mutex fixtures, thread-map bands test); Verilator 5.048 lock-step still cycle-exact on the preemption fixture. The only failures are the two pre-existing `construct_proof_lean_*` local Lean-toolchain env failures (fail on unmodified base `f028f02`).

nogate-ai:
REVIEW VERDICT: APPROVE — no blocking issues found

Independent review performed from a fresh `cargo build --release` on this worktree's checkout of `claude/lock-arbiter-hold-stability` (commit `ad3ddb9`). Reviewed the full diff (`gh pr diff 691`), read the changed source in context (not just the diff hunks), and ran the suggested targeted test slice plus a full local `cargo test --release`.

## What I verified

**Bug 1 (SV hold latch, `src/codegen/arbiter.rs` + `src/elaborate.rs`)**
- `hold_valid_r`/`hold_owner_r` gate is applied uniformly across all policy emitters (round_robin, priority, custom-hook, and LRU/weighted which reuse the round_robin/priority emitters respectively) — traced `emit_arbiter`'s match arms, all pass `a.lock_hold` through.
- `hold_valid_r <= grant_valid && !release[grant_requester]` / `hold_owner_r <= grant_requester` (on `grant_valid`) is correct: request-deassertion re-arbitrates combinationally the same cycle (owner check is gated on `request_valid[hold_owner_r]`), while the `release` pulse clears the hold register one cycle later for back-to-back re-lock (request never drops) — matches the spec text added in `doc/ARCH_HDL_Specification.md` / `thread_spec_section.md` describing same-cycle vs. next-cycle handoff for the two cases.
- `rr_ptr_r` continues to update every held cycle to `hold_owner_r + 1`, which is idempotent while held and correctly leaves the scan pointer at "one past the outgoing owner" for the next free-resource arbitration — matches the round-robin/LRU doc section's claim.
- `lock_hold` is only ever set `true` in `synthesize_lock_arbiter`; the parser's own `ArbiterDecl` construction defaults it `false`. Ran `test_lock_arbiter_emits_hold_latch_user_arbiter_does_not` — confirms synthesized lock arbiters carry `hold_valid_r`/`request_release` and user-declared `arbiter` constructs carry neither.
- arch-sim's C++ arbiter model (`src/sim_codegen/mod.rs`, `_hold_valid`/`_hold_owner`) mirrors the SV FF/comb shape 1:1, including the `!grant_valid` guard added to the round-robin scan loop so the hold check pre-empts the scan (previously `if ((valid>>idx)&1)` unconditionally, now `if (!grant_valid && ...)`).

**Bug 1 follow-up (`request_release` / `lock_release` fold, `src/elaborate.rs`)**
- Traced all five `lock_release` fire-condition branches (`is_folded`, `multi_transitions`, `transition_cond`, `wait_cycles`, unconditional). The `is_folded` branch reads `raw_states[si-1].transition_cond`, which `fold_wait_until_exit_assignments` only sets `is_folded` on a successor when the predecessor's `transition_cond.is_some()` was already verified, so the `.clone()` is never used against `None` in that path.
- The `wait_cycles` branch's synthesized fire condition (`_t{ti}_cnt == 0`) uses the same counter-naming convention (`_t{ti}_cnt`) used everywhere else in `lower_module_threads` for per-thread wait counters — verified by grep, not just assumed.
- Default-0 assignment for `_<res>_release_<ti>` is emitted alongside the existing default-0 for `_<res>_req_<ti>`, so unmarked states correctly never pulse release.

**Bug 2 (`eval_comb()` internal settle, `src/sim_codegen/mod.rs`)**
- `comb_settle_depth = if !insts.is_empty() { settle_depth } else { 1 }` correctly leaves modules with no sub-instances (and thus `settle_depth == 1` modules) emitting a single pass — no behavior change for the common case.
- One non-blocking observation (see Nits) about how this composes with the *caller's own* settle loop.

**Bug 3 (coroutine scheduler, `runtime/arch_thread_rt.h` + `src/sim_codegen/thread_sim.rs`)**
- Traced the `tick_begin()`/`tick_run()` split and the `lock_wait`/`lock_refired` one-shot re-fire logic by hand against several interleavings: (a) a slot freshly resumed this tick that immediately re-suspends on `lock_wait_until` and finds the resource free — gets exactly one bonus same-tick re-fire, consistent with "zero-latency acquisition of a free resource"; (b) a slot already `WaitUntil` from a prior tick (not lock-related) — unaffected, still resumes via the ordinary `!resumed[i]` path with no double-resume; (c) a spurious immediate re-suspend on `lock_wait_until` after the one-shot re-fire is consumed — correctly blocked from re-firing again the same tick because `lock_refired[i]` stays set for the rest of the tick while `resumed[i]` is also true again post re-fire, so `eligible` evaluates false. Regular (non-lock) `wait_until` explicitly passes `lock_acquire=false`, so it's structurally impossible for a normal wait to hit the re-fire branch.
- `gen_module_thread_with_warnings` correctly switched all three call sites (initial settle in the constructor, reset-triggered re-settle, and the normal sequential tick) from per-scheduler `tick()` to `tick_begin()` + a cross-scheduler fixed-point loop over `tick_run()`.

**Semaphore interaction (deferred hazard, issue #696)**
- Confirmed directly in `elaborate.rs`: for `n_slots > 1`, the packed `request_valid` fed to the arbiter is `waiting_i = req_i & !held_i`, not raw `req_i`. Once a thread is admitted, `held_i` latches and `waiting_i` drops combinationally the same cycle, so the arbiter's `hold_valid_r`/`hold_owner_r` for that thread self-clears on the very next arbitration (owner's `request_valid` bit is 0) — matches the PR's claim that the hold latch is harmless for semaphores. `request_release` is packed and connected for all resource kinds but the semaphore `held_i` register logic doesn't consume it — this is explicitly called out in both the PR body and inline comments as tracked separately (#696), not silently left as a surprise.

**Docs**
- `doc/thread_lowering_proof.md` Lemma L, `doc/thread_lowering_algorithm.md` (hold latch, liveness-per-policy, nested-lock rationale), `doc/thread_spec_section.md` §20.8.3, `doc/ARCH_HDL_Specification.md` §7a.3, and `doc/plan_thread_parallel_sim.md`'s new timing-contract section are all consistent with what the code actually does — no doc/implementation drift found. `skills/arch-programming/references/ARCH_HDL_Specification.md` mirrors the canonical spec edit verbatim (skill-snapshot in sync, confirmed by CI too).

## Testing performed

- `cargo build --release` fresh from this worktree (verified no stale-source warnings).
- `cargo test --release --test integration_test -- mutex lock_arbiter_emits semaphore axi_read_beat_interleave resource_lock`: **24/24 pass**, including all four new mutex fixtures (`mutex_release_rr`, `mutex_release_pri`, `mutex_hold_preempt_pri`, `mutex_hold_preempt_rr`), the beat-interleave regression (arch-sim + Verilator variants), and the hold-latch/user-arbiter golden test.
- Full local `cargo test --release`: only failures are the 2 pre-existing `construct_proof_lean_*` local Lean-toolchain-path tests, unrelated to this diff (matches the PR's own disclosure and this repo's known recurring local-elan-path issue). `gh pr checks 691` shows all CI jobs green, including `cargo test`, which doesn't hit this local-environment issue.

## Nits (non-blocking)

1. **`eval_comb()` self-settle nests inside the caller's own settle loop (`src/sim_codegen/mod.rs`).** For a module with sub-instances, `eval()`'s own settle loop iterates `settle_depth` times and, on each iteration, calls this module's own `eval_comb()` ("parent comb within settle loop", around the `// Parent comb within settle loop` comment) — but `eval_comb()` itself now *also* loops `comb_settle_depth` (== the same `settle_depth`) times internally, and each of those inner iterations re-invokes every child instance's `eval_comb()` again. Net effect: for a module with sub-instances, one call to `eval()` does on the order of `settle_depth²` comb passes instead of `settle_depth`, and this compounds multiplicatively through each additional level of instantiation depth (each child's own `eval_comb()` may itself wrap further nested instances). Since ARCH comb blocks are required to be pure functions of current state, the extra iterations are fixed-point no-ops and not a correctness bug, but it's an avoidable simulation-performance regression worth a follow-up for deeply hierarchical designs (e.g. multi-level bus/arbiter trees). A cleaner split would be: keep `eval_comb()` self-settling for callers that invoke it as a leaf (e.g. a parent's inst-settle loop calling a *child's* `eval_comb()`), but have the module's own `eval()` call an internal single-pass comb helper rather than the public self-settling `eval_comb()`, so the two loops don't multiply.
2. Everything else — naming, span/thread-map handling in the `lock_release` injection, and the doc/spec updates — was in-scope, well-commented, and matched the intent described in the PR body.

No correctness, mutual-exclusion, or liveness issues found in the hold latch, the release-pulse synthesis, the sim/coroutine backends, or the semaphore interaction. Recommend addressing the settle-loop nit as a low-priority follow-up rather than blocking this PR on it.

PR #695, opened 2026-07-13
feat(lang): compile-time param where constraints (closes #600)

### LINKED ISSUE #600: Enhancement: compile-time parameter constraint `where` clauses
## Problem

ARCH modules are routinely parameterized over quantities like cache depths, FIFO sizes, bus widths, and pipeline stage counts. Nearly all such parameters carry implicit validity constraints — a cache size must be a power of two for address decoding, a FIFO depth must be positive, a pipeline must have at least one stage, a bus width must be a power of two and ≥ 8. Today there is no way to express these constraints **in the module declaration itself**.

```arch
// Current state: constraints live only in comments
module Cache
  param SIZE:   const = 1024;   // must be a power of two and > 0
  param DATA_W: const = 32;     // must be 8, 16, 32, or 64
  param STAGES: const = 2;      // must be >= 1
```

Violations are discovered only when the flawed parameters reach SV synthesis (cryptic width mismatch), simulation (array out-of-bounds abort), or formal (vacuous proof on a zero-depth structure). None of those sites point back to the instantiation that was wrong.

This gap is especially sharp for ARCH's primary use case: **LLM-generated hardware**. A model generates a cache instantiation with `SIZE=3` or `STAGES=0` and every downstream tool fails silently or with a generic error that gives no hint about the constraint violation.

---

## Proposed syntax

Add an optional `where` clause to `param` declarations naming a compile-time Boolean predicate:

```arch
module Cache
  param SIZE:   const = 1024 where SIZE > 0 and (SIZE & (SIZE - 1)) == 0;
  param DATA_W: const = 32   where DATA_W == 8 or DATA_W == 16 or DATA_W == 32 or DATA_W == 64;
  param STAGES: const = 2    where STAGES >= 1 and STAGES <= 8;
  ...
end module Cache
```

Grammar extension (minimal — one new optional tail on existing `ParamDecl`):

```
ParamDecl := 'param' IDENT ':' ParamKind '=' ConstExpr ('where' ConstExpr)? ';'
```

The `where` expression:
- Is a constant Boolean expression evaluated at instantiation time, same rules as param defaults (only literals, earlier `const` params, and compile-time intrinsics like `$clog2`).
- May reference the declared parameter by name and any earlier `const` params in the same block.
- Must type-check as `Bool`; a non-Bool is a compile error at the definition site.
- `where` is currently an unused keyword in ARCH — no collision.

---

### 1 — Definition-site validation

After type-checking the module, const-evaluate the `where` expression against the declared default. A violated default is a compile error:

```
error: default value SIZE=3 violates constraint: (SIZE & (SIZE - 1)) == 0
  --> Cache.arch:2
```

### 2 — Instantiation-site enforcement

At every `inst ... param NAME = VALUE;` site, `arch check`/`build`/`sim`/`formal` evaluates the constraint with the concrete value and reports:

```
error: param SIZE=3 in inst l1d violates constraint: (SIZE & (SIZE - 1)) == 0
       (SIZE must be a power of two and > 0)
  --> top.arch:14:17
   |
14 |   inst l1d: Cache param SIZE = 3;
   |                          ^^^ evaluated here
note: constraint declared at Cache.arch:2: SIZE > 0 and (SIZE & (SIZE - 1)) == 0
```

### 3 — Separate compilation

The `where` clause is emitted verbatim in `.archi` interface files so downstream compilation units enforce the constraint at their own instantiation sites.

### 4 — No codegen impact

Constraints are erased after the type-check pass. The emitted SV, SMT-LIB2, and sim C++ are identical to today's output.

---

## Why this matters for LLM-generated ARCH

When an LLM reads a module header, it now sees the constraint alongside the parameter name:

```arch
param SIZE: const = 1024 where SIZE > 0 and (SIZE & (SIZE - 1)) == 0;
```

The model can self-validate before emitting an instantiation — `SIZE=512` visibly passes, `SIZE=300` visibly fails. When the model gets it wrong anyway, `arch check` delivers a clear, actionable error at the instantiation site rather than a cryptic failure 3 passes later. The `arch learn` system records `constraint-violation → fix` pairs exactly like any other error class, improving future suggestions via `arch advise`.

---

## Relation to existing features

| Feature | Scope | Fires when |
|---|---|---|
| `where` clause (proposed) | Compile time | `arch check`, always |
| `assert name: expr;` | Runtime / formal | `arch sim`, `arch formal` |
| `generate if PARAM > 0 ...` | Conditional codegen | Selects paths; never errors |
| Divide-by-zero in const-eval | Compile time | Only `A/0` in constant expressions |

The proposed `where` clause fills the gap: user-declared compile-time rejection of invalid parameter values. It is complementary to `assert` (which runs in simulation) and `generate if` (which routes around, not rejects).

---

## Implementation sketch

All changes are in `src/parser.rs`, `src/ast.rs`, and `src/typecheck.rs`. No codegen files are touched.

### AST change (`src/ast.rs`)

```rust
pub struct ParamDecl {
    pub name:          Ident,
    pub kind:          ParamKind,
    pub default:       Option<Expr>,
    pub constraint:    Option<Expr>,   // ← new: the `where` expression
    pub is_local:      bool,
    pub span:          Span,
    pub unpacked_size: Option<Expr>,
}
```

### Parser change (`src/parser.rs`)

After parsing the default value in `parse_param_decl`, check for `where` and, if present, parse the following `ConstExpr` into `constraint`.

### Type-checker changes (`src/typecheck.rs`)

**Definition site**: if `constraint` is `Some(expr)`:
1. Type-check `expr` with the param name bound to its declared type; require result type `Bool`.
2. Const-evaluate against the default; if `false`, emit definition-site error.

**Instantiation site**: when processing `param NAME = VALUE` in an `inst` block:
1. Look up the module's `ParamDecl` for `NAME`.
2. If `constraint` is `Some(expr)`, const-evaluate with `NAME` bound to `VALUE`.
3. If `false`, emit the targeted error with both `inst` span and definition span.

**`.archi` emitter**: serialize `constraint` alongside the existing param fields. On import, deserialize; the same instantiation-site check runs unchanged.

---

## Staged rollout

**v1 — single-param `where`**: `where` can only reference the declared param itself. Covers the most common cases (range checks, power-of-two, bit-width alignment).

**v2 — cross-param `where`**: allow references to any earlier `const` param in the same block:

```arch
param DATA_W: const = 32  where DATA_W >= 8 and DATA_W % 8 == 0;
param STRB_W: const = 4   where STRB_W == DATA_W / 8;
```

The const-eval pass already resolves earlier params before later ones at instantiation time, so the machinery is already there — just broaden the allowed reference set.

**v3 — built-in predicates** (optional convenience):

```arch
param SIZE: const = 1024 where is_pow2(SIZE) and SIZE > 0;
```

`is_pow2` as a compile-time intrinsic avoids the unreadable `(SIZE & (SIZE - 1)) == 0` idiom. Other candidates: `is_aligned(x, n)`, `in_range(x, lo, hi)`. Defer until usage shows which ones are worth reserving.

---

## Non-goals for v1

- **Type-param constraints** (`param T: type = UInt<8> where bitsof(T) >= 8`). Requires a `bitsof()`/`widthof()` intrinsic that doesn't exist yet.
- **Runtime enforcement**. `where` constraints are compile-time only. ARCH has no dynamic instantiation.
- **Cross-module named predicates** (`constraint IS_POW2 = ...`). Useful for reducing repetition, but anonymous inline `where` ships first.

---

## Prior art

- **Ada**: range types `type Index is 0 .. 255` with compiler-enforced bounds — closest analogue.
- **Chisel**: `require(SIZE > 0 && isPow2(SIZE), "SIZE must be a power-of-two")` — Scala runtime check in module body; fires at elaboration, not at parameter declaration.
- **SystemVerilog**: no equivalent. `parameter` declarations have no constraint mechanism; constraints live in comments or in separate documentation.
- **Clash / Haskell**: type-level natural number constraints via `KnownNat`; more powerful but also far heavier than what ARCH needs.

ARCH's `where` clause takes the Ada approach (declarative, inline with the parameter) without requiring a full type-level constraint solver.

nogate-ai:
## Summary

Implements issue #600 (maintainer-approved v1 minimal scope): optional compile-time `where` constraint clauses on `const` params.

```arch
module Cache
  param SIZE:   const = 1024 where SIZE > 0 and (SIZE & (SIZE - 1)) == 0;
  param DATA_W: const = 32   where DATA_W == 8 or DATA_W == 16 or DATA_W == 32 or DATA_W == 64;
  param STAGES: const = 2    where STAGES >= 1 and STAGES <= 8;
```

closes #600

## What landed

**Grammar** — `ParamDecl := 'param' IDENT ':' ParamKind '=' ConstExpr ('where' ConstExpr)? ';'`. `where` is a contextual keyword (no lexer reservation needed — verified unused). Only legal on `const` / width-qualified const params; a `where` on a type/logic/enum/vec param is a parse error: `` `where` constraint clauses are only allowed on `const` params, not type/logic/enum/vec params ``.

**Definition-site check** (`src/typecheck.rs::check_param_where_default`) — constraint must type-check as `Bool` (error: `` param `N` where clause must be a Bool expression, found UInt(4) ``), then is const-evaluated against the declared default. Violation:
```
error: default value SIZE=3 violates constraint: ((SIZE > 0) and ((SIZE & (SIZE - 1)) == 0))
```
The where-expr may reference the declared param itself and earlier const params in the same block; `$clog2` etc. are supported via the shared const-eval (`elaborate::try_eval_i64`).

**Instantiation-site check** (`src/typecheck.rs::check_inst_param_constraints`) — at every `inst ... param NAME = VALUE;` override, constraints are re-evaluated with effective (post-override) values built in declaration order, so derived-param chains (`param B: const = A*2 where B <= 64;` with `param A = 40` overridden) are caught:
```
error: param `SIZE`=3 in inst `l1d` violates constraint declared on `Cache`: ((SIZE > 0) and ((SIZE & (SIZE - 1)) == 0))
```
Primary span is the override value expression at the inst site; the message quotes the constraint source verbatim (no message-string syntax in v1). generate_for-computed params are enforced via the monomorphized-variant path (elaboration bakes the computed value into the variant's default, which the definition-site check then rejects) — tested.

**.archi propagation** — `src/interface.rs` emits `where <expr>` verbatim for Const/WidthConst params; downstream compilation units enforce at their own inst sites (tested end-to-end with the source file deleted so only the `.archi` stub is visible).

**Erased before codegen** — constraint lives only in the AST + typecheck; no codegen file touched. SV output tested byte-identical with/without satisfied clauses.

**Constructs covered** — all constructs sharing `ParamDecl`: module, fsm, fifo, ram, cam, arbiter, counter, regfile, pipeline, synchronizer, clkgate, template, linklist, bus.

**All commands enforce** — the check runs in the shared typecheck pass used by `arch check` / `build` / `sim` / `formal`.

## Docs (same PR)

- `doc/ARCH_HDL_Specification.md` — new §4.1a (grammar, evaluation rules, both check sites, .archi + erasure notes) + schema-line mention in §4.1
- `doc/Arch_AI_Reference_Card.md` — schema row with power-of-two example
- `doc/arch.ebnf` — `param_decl` production updated
- `skills/arch-programming/references/` snapshots refreshed via `scripts/sync_skill_snapshots.sh refresh` (check passes)
- `mcp/instructions.md` — inspected; teaches construct selection, not param grammar — no change needed

## Tests

`tests/param_where_constraints.rs` — 11 tests: violated default (message locked) / satisfied default / non-Bool where / where-on-type-param rejected / violated override (site + declaring construct + verbatim constraint) / satisfied override / generate_for computed violation / derived-param violated + satisfied / .archi separate compilation (real binary) / SV codegen purity. Includes the end-to-end Cache SIZE/DATA_W/STAGES fixture from the issue.

## Verification (fast gate, all green before opening)

- `cargo build --release` clean
- `cargo test --release` — 870 tests, 0 failures (65 lib + 23 formal + 23 sim + 731 integration + 11 new + 12 + 3 + 2)
- `cargo fmt --check` clean
- `cargo clippy --release --all-targets` — no new warnings from this change (one introduced `redundant_guards` was fixed)
- Lean `lake build` under `proofs/lean_thread_lowering` (fresh-worktree artifact) rebuilt; the two Lean replay tests pass


nogate-ai:
Follow-up commit 5c61645: bus params are overridden at port declarations (`port b: initiator MiniBus<DATA_W=48>;`), not inst sites — added the bus analogue of the inst-site check (`check_bus_port_param_constraints`) so bus where-clauses are enforced end-to-end. Error format: `param \`DATA_W\`=48 on bus port \`b\` violates constraint declared on \`MiniBus\`: ...`. Spec §4.1a updated, snapshots refreshed, 3 new tests (14 total). Full suite re-run green locally (873 tests).

PR #697, opened 2026-07-13
refactor(sim): deduplicate FP param const eval into helper

nogate-ai:
Follow-up to #694: the FP param emission logic was duplicated in two places (package/module header and module class header). This refactors it into eval_param_const_value helper, addressing the tech debt noted in #694 review.

- Extracts FP and int param evaluation into single helper
- Handles both FP32/BF16 via f64_to_fp32/bf16_bits
- Used in both emission sites

Verified: cargo test PASS, Bf16ToFixed both sims PASS


nogate-ai:
Superseded by #700 which includes helper refactor

PR #698, opened 2026-07-13
fix/fp param helper2

nogate-ai:
- **fix(fp): wire FP types not recognized in SV codegen**
- **refactor(sim): deduplicate FP param const eval into helper**


nogate-ai:
Superseded by #700

PR #699, opened 2026-07-13
refactor(sim): deduplicate FP param const eval into helper

nogate-ai:
Follow-up to #693 and #694: duplicated FP param emission logic.

- Extract eval_param_const_value() helper
- Fix type_bits_te_with_params BF16 width and param_float_fmt FSM handling
- Addresses review feedback from #694


nogate-ai:
Superseded by #700

PR #701, opened 2026-07-13
perf(sim): eval() delegates comb settling to self-settling eval_comb(); fix settle_depth undercount it exposed

nogate-ai:
## What

Follow-up to the non-blocking perf note on #691: for modules with sub-instances, `eval()` ran its own settle loop calling the (now self-settling) `eval_comb()` once per pass — O(settle_depth²) comb evaluations of the same module per `eval()`, compounding through instantiation depth.

`eval()` now delegates entirely, matching the no-inst path:

```cpp
eval_comb();      // self-settling: inst chain + parent comb, settle_depth passes
eval_posedge();   // edge detect + parent regs + child posedges (unchanged)
eval_comb();      // post-posedge re-settle
```

The duplicated inst input-mirroring / output-reading chains in `eval()` are deleted (~880 lines of emitter code). `eval_comb()`'s chain is the single canonical copy, and it now:
- iterates in the topological `inst_eval_order` the settle analysis assumes (was source order — previously only safe because `eval()`'s topo-ordered loop ran too), and
- carries the `--check-uninit` vinit marking that previously existed only in `eval()`'s copy.

## The undercount the O(d²) loop was hiding

The review premise that the double loop was *correctness-neutral* turned out to be false. `comb_graph::analyze_module` undercounted `settle_depth` by one pass for an inst output consumed through an intermediate produced by a **later-in-source** comb block — exactly the lock lowering's pack/unpack bridge: the thread comb reads `_grant_i`, which the bridge block unpacks from the arbiter's `grant_packed` *after* the thread comb runs. The chain

```
state ─▸ req_i ─▸ req_packed ─▸ arbiter ─▸ grant_packed ─▸ grant_i ─▸ busy
        (pass p)  (tail of p)   (pass p+1)                (tail p+1)  (pass p+2)
```

needs **3** passes; the analysis said 2. The old d×d = 4 effective passes masked it. With delegation alone, `busy` lagged one cycle vs baseline — the exact grant-lag infidelity #691 fixed. `settle_depth` is now 3 (acyclic with parent comb intermediates) / 4 (cyclic with intermediates), with the per-pass model documented at step 6, and the pinned unit test updated.

Internal-only change: no user-facing syntax/semantics touched; generated sim behavior is trace-identical (below).

## Verification

- `cargo test --release --no-fail-fast`: all suites green — 736/736 integration (incl. `mutex_release_rr/pri`, `mutex_hold_preempt_*`, `axi_read_beat_interleave`, thread-sim-both cross-checks). The 2 `formal_test` Lean failures are the known pre-existing environment issue (fail identically on base commit f08827f).
- `--debug` trace diff, freshly built base (f08827f) vs this branch, over all 16 `arch_sim_manifest` entries + the four mutex fixtures + `ThreadAxiReadBeatInterleave`: **byte-identical traces**. (One manifest entry, `axi_dma_tlm__TlmMm2sBurstVec`, fails identically on both binaries — pre-existing `--debug` codegen bug for wide TLM ports, unrelated.)
- `--thread-sim both` cross-check on mutex fixtures: `Cross-check PASS`.
- Long verification (full regression sweep + harc equivalence harness with `ARCH_BIN`) running post-open per the fast-gate-before-PR flow; follow-ups will be pushed if it surfaces anything.


nogate-ai:
Long verification complete, all green:
- `tools/run_arch_regression.py` (fresh branch binary, vs `tests/arch_regression_baseline.json`): **179/179 pass**.
- harc `tests/run_tbir_equiv.sh` with `ARCH_BIN` pointing at this branch's binary (v1-vs-tbir trace-diff incl. the ARCH-native-DUT sweep rows): **174/174 pass, 0 failed**.

No follow-up commits needed.

PR #703, opened 2026-07-14
fix(threads): free semaphore slot on lock-release pulse (closes #696)

### LINKED ISSUE #696: semaphore<N>: back-to-back re-lock never releases its slot (held_i keyed on req deassert)
Found while merging the semaphore lowering (#689) with the lock-arbiter hold-stability fix (PR #691).

For `n_slots > 1`, the semaphore holder register releases only when the holder's request wire deasserts:

```
held_i <= admit_i ? 1 : (held_i & req_i ? held_i : 0)
```

A thread that releases and immediately re-locks the same resource (back-to-back `lock` in a loop — the beat-interleaving idiom) never deasserts `req_i` across the `end lock` boundary, so `held_i` stays 1 across acquisitions: the thread retains its slot forever. With more waiters than slots, waiters starve, and per-boundary rotation never happens — exactly the mutex-side bug PR #691 fixed with the per-thread `request_release[i]` pulse.

The fix is small now that #691's plumbing exists: the merged module already packs `_<res>_release_packed` (driven by each lock's last body state on its exit transition) for every resource kind. The semaphore held-register update just needs to consume it:

```
held_i <= admit_i ? 1 : (held_i & req_i & !release_i ? held_i : 0)
```

plus a regression fixture (2 slots, 3 lanes, per-beat lock/release loop → expect rotation) mirroring `tests/thread/mutex_release_rr.arch`. Note the re-releasing thread must still be able to re-acquire the same cycle if a slot stays free (`admit_i` path), so a lone holder keeps streaming with no bubble.

Deliberately not folded into #691 to avoid changing just-merged semaphore semantics without its own review + fixtures.

nogate-ai:
## Summary

Closes #696. A `semaphore<N, policy>` tracks each thread's slot in a per-thread `held_i` register whose next-state was:

```
held_i <= admit_i ? 1 : (held_i & req_i)
```

That frees a slot only when the request wire deasserts. But a thread that **re-locks the same resource back-to-back** — a tight loop where `end lock` is immediately followed by the next `lock` with no intervening wait — never deasserts `req_i` across the boundary, so `held_i` stays `1` forever. With `semaphore<2>` and three contenders, the two slots pin on T0/T1 and **T2 is starved**.

The end-of-lock-body release pulse `_<res>_release_<ti>` (added in #691, already consumed by the *mutex* arbiter hold-latch via `request_release`) was never wired into the *semaphore* `held_i` next-state — even though the release-pulse generator's own comment already states the held register should clear on the pulse. The fix gates the hold on `!release_i`:

```
held_i <= admit_i ? 1 : (held_i & req_i & !release_i)
```

## Why this is more than a timing nit

It is a **miscompile of the synthesizable RTL**. The FSM-lowered backend — what `arch build` emits and what default `arch sim` runs — silently **diverged** from the coroutine `--thread-sim parallel` backend (which rotated fairly). On a 3-lane `semaphore<2, round_robin>` tight loop:

| backend | before fix | after fix |
|---|---|---|
| FSM-lowered (`arch build` / default sim) | `cnt0=301 cnt1=300 cnt2=0` (T2 starved) | `cnt0=201 cnt1=200 cnt2=200` |
| coroutine (`--thread-sim parallel`) | `cnt0=202 cnt1=201 cnt2=201` (fair) | fair |

Emitted SV after the fix:
```sv
_pool_held_0 <= _pool_admit_0 ? 1 : _pool_held_0 && _pool_req_0 && !_pool_release_0;
```

## Scope

Internal lowering fix (`src/elaborate.rs`) only — no user-facing syntax or semantics change, so no grammar/doc surface is touched.

## Regression coverage

- **`test_semaphore_slot_freed_by_release_pulse_696`** (structural): the `held` next-state must gate on `!_pool_release_<ti>`.
- **`test_semaphore_relock_round_robin_fairness_696`** (runtime, FSM backend): three tight-re-lock lanes on `semaphore<2, round_robin>` must all make balanced progress (new fixture `tests/thread/semaphore_relock_rr.arch` + TB). This also closes the coverage gap that every prior both-backend semaphore/mutex fixture used `thread once` and **never re-locked** — which is exactly how the divergence shipped green.

## Verification

- Fresh `cargo build --release` + full `cargo test --release`: green (the only failures are the two pre-existing Lean-formal env tests that need `lake`/`elan` on PATH — unrelated to this change; they fail identically on `origin/main`).
- New tests pass; all existing `mutex`/`semaphore`/`lock` tests pass.
- Manual FSM-vs-coroutine repro table above, verified on the fresh binary.

## Follow-up (separate, not fixed here)

- The two backends still differ by a first-cycle **reset-timing** offset on a *gate-less* tight loop (coroutine increments at cycle 0 during reset; FSM waits for reset to deassert), so `--thread-sim both` cross-check flags event #1 on this fixture. That is orthogonal to the slot leak and predates this PR — flagging for a separate look.
- `arch check` emits a whole-design comb-SCC warning on the tight re-lock idiom (the arbiter grant↔release feedback, registered through the hold-latch so likely a conservative false positive). Also pre-existing to this fix; noted for follow-up.

### DIFF
--- a/src/elaborate.rs
+++ b/src/elaborate.rs
@@ -4667,11 +4667,37 @@ fn lower_module_threads(
                     Expr::new(ExprKind::Ident(format!("_{}_admit_{}", res_name, ti)), sp);
                 let held_ident = Expr::new(ExprKind::Ident(held_name.clone()), sp);
                 let req_ident = Expr::new(ExprKind::Ident(format!("_{}_req_{}", res_name, ti)), sp);
-                let still_held = Expr::new(
+                // #696: a semaphore slot is freed by the end-of-lock-body release
+                // pulse (`_<res>_release_<ti>`), NOT only by the request wire
+                // deasserting. A thread that re-locks back-to-back (tight loop)
+                // never deasserts `req_i` across `end lock`, so `held_i & req_i`
+                // alone would pin the slot forever and starve waiting contenders.
+                // Gate the hold on `!release_i` so the slot rotates on the pulse,
+                // matching the arbiter hold-latch (which already consumes the same
+                // pulse via `request_release`, see lock-release-pulse generation).
+                let not_release = Expr::new(
+                    ExprKind::Unary(
+                        UnaryOp::Not,
+                        Box::new(Expr::new(
+                            ExprKind::Ident(format!("_{}_release_{}", res_name, ti)),
+                            sp,
+                        )),
+                    ),
+                    sp,
+                );
+                let held_and_req = Expr::new(
                     ExprKind::Binary(
                         BinOp::And,
                         Box::new(held_ident.clone()),
                         Box::new(req_ident),
+                    ),
+                    sp,
+                );
+                let still_held = Expr::new(
+                    ExprKind::Binary(
+                        BinOp::And,
+                        Box::new(held_and_req),
+                        Box::new(not_release),
                     ),
                     sp,
                 );
--- a/src/elaborate.rs
+++ b/src/elaborate.rs
@@ -4694,11 +4694,7 @@ fn lower_module_threads(
                     sp,
                 );
                 let still_held = Expr::new(
-                    ExprKind::Binary(
-                        BinOp::And,
-                        Box::new(held_and_req),
-                        Box::new(not_release),
-                    ),
+                    ExprKind::Binary(BinOp::And, Box::new(held_and_req), Box::new(not_release)),
                     sp,
                 );
                 let next_val = Expr::new(
--- a/tests/integration_test.rs
+++ b/tests/integration_test.rs
@@ -17863,6 +17863,40 @@ fn test_semaphore_relock_round_robin_fairness_696() {
     );
 }
 
+// arch#696 regression (runtime, safety bound): the SAME tight-re-lock churn must
+// never over-subscribe the pool. Four lanes hold a semaphore<2> until their
+// TB-driven release; the TB asserts busy<=2 EVERY cycle, that the pool actually
+// reaches 2 concurrent holders, and drives 60 rounds of release+re-admit churn —
+// the release path the fix newly exercises. Guards against the fix freeing/
+// re-admitting a slot such that >N threads hold at once (adversarial-review gap:
+// the fairness test alone did not check the concurrency bound).
+#[test]
+fn test_semaphore_relock_concurrency_bound_696() {
+    let td = tempfile::tempdir().expect("tempdir");
+    let arch_bin = env!("CARGO_BIN_EXE_arch");
+    let out = std::process::Command::new(arch_bin)
+        .arg("sim")
+        .arg("tests/thread/semaphore_relock_bound.arch")
+        .arg("--tb")
+        .arg("tests/thread/tb_semaphore_relock_bound.cpp")
+        .arg("--outdir")
+        .arg(td.path())
+        .output()
+        .expect("run arch sim on semaphore_relock_bound");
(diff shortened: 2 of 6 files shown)


nogate-ai:
### Long verification: real Verilator 5.048 on the emitted SV

Confirmed the fix at the synthesizable-RTL level (not just `arch sim`), and confirmed Verilator *catches* the pre-fix bug:

| emitted SV `held` next-state | Verilator 5.048 result |
|---|---|
| `admit ? 1 : held & req` (pre-fix) | `cnt0=300 cnt1=299 cnt2=0` → **starvation** |
| `admit ? 1 : held & req & !release` (this PR) | `cnt0=200 cnt1=200 cnt2=199` → **all lanes progress** |

Method: `arch build semaphore_relock_rr.arch -o sem.sv`, then `verilator --cc --exe --build --top-module SemaphoreRelock sem.sv tb.cpp` (300-cycle C++ harness). The pre-fix column was produced by `sed`-stripping the `&& !_pool_release_<n>` term from the same emitted SV, isolating the one-term change as the sole cause. Full `cargo test --release` also green (only the two pre-existing Lean-env tests fail, identically on `origin/main`).

nogate-ai:
### Adversarial review (self-review, attempting to refute the fix)

Tried to break the fix along four axes. It holds — but the pass surfaced one real coverage gap (now fixed) and one mischaracterization in this PR's own notes (corrected below).

**Safety — over-subscription (>N holders):** the admit gate is a single scalar `holder_count < N` shared across threads, so I checked whether the release/re-admit churn my fix enables could admit past a free slot. It cannot: `holder_count = popcount(held_*)` and the arbiter emits one `grant_requester`, so `grant_packed` is one-hot ⇒ ≤1 admit/cycle ⇒ the bound holds by construction. Confirmed empirically — a 4-lane `semaphore<2>` under 60 rounds of release/re-admit churn reaches exactly 2 and **never exceeds 2**, on FSM sim, coroutine sim, and **Verilator on the emitted SV**.

**Safety — mutual exclusion for `semaphore<1>`≡`mutex`:** the `held_i` path also lowers mutex, so I ran the same probe as a tight-re-lock `mutex<round_robin>`: max concurrent = 1, never exceeded, both backends. Mutual exclusion preserved.

**Liveness — under-utilization / lost slot:** confirmed the pool still reaches N and the releaser rotates back in (no slot stranded). Release fires only on the last lock-body state's exit transition, so `held_i` stays asserted for the entire hold and drops exactly at exit — not mid-critical-section.

**Coverage gap found & closed:** the original regression checked *fairness* but never the *concurrency bound*. Added `test_semaphore_relock_concurrency_bound_696` (+ fixture `semaphore_relock_bound.arch`) asserting busy≤N every cycle and that N is reached, through churn. Pushed in 21aec24.

**Correction to this PR's follow-up note:** I wrote the `arch check` comb-SCC warning was "likely a conservative false positive." That is **wrong** — Verilator 5.048 confirms it as a genuine `UNOPTFLAT` (circular combinational logic) on `_pool_grant_*` in the emitted SV. It is **pre-existing** (from #691's zero-cycle release plumbing, not this fix — verified by stripping the `!release` term from the SV: the loop persists, and never involves `_pool_held`), and it is a real synthesizability hazard for the tight-lock idiom. Filed as #709 with full isolation evidence.

Net: the fix is correct on all safety/liveness/timing axes; regression strengthened; the one inaccurate note is corrected and its underlying (pre-existing) issue tracked separately.

PR #706, opened 2026-07-14
fix(lexer): remove redundant alternative in SizedLiteral regex

nogate-ai:
Second alternative [0-9]+[a-zA-Z_][a-zA-Z0-9_]*'[bhd][0-9a-fA-F_]+
is unreachable since ARCH param names can't start with digit.
First alternative already covers W'd0, third covers 8'hAB etc.

Fixes remaining-findings.md low: Lexer regex has redundant alternative


nogate-ai:
Superseded by #707 which includes lexer fix + clog2 handling

PR #735, opened 2026-07-30
Rewrite cocotb shim: event-driven ps scheduler, full trigger surface, unmodified cocotbext-axi support

teabone113:
## Summary

Implements the **ArchNativeCocotbShim** spec end to end: ordinary cocotb tests and the installed `cocotbext-axi` package now drive `arch sim --pybind --test` native models directly.

### Scheduler semantics
- **Integer-picosecond time base** shared by `Clock`/`Timer`/timeouts/`get_sim_time()`; a 3333 ps clock alternates exact 1666/1667 ps half-periods.
- **Event-driven advancement** — time jumps to the next timer/clock deadline; no per-nanosecond model evaluation.
- **`ReadOnly()` is a scheduler phase transition** in the current timestep (previously a next-tick zero-duration timer).
- **Region-ordered timesteps**: `.value =` writes are buffered deposits (cocotb NBA-style; `setimmediatevalue()` for immediate), edge waiters resume *before* the edge evaluates and observe pre-edge state, reaction writes land after. This ordering is what cocotbext-axi's ready/valid source models depend on — without it, handshakes double-accept.
- Deadlock detection with named waited-on signals instead of silent hangs.

### API surface
`Edge`, `ReadOnly`, `ReadWrite`, `NextTimeStep`, `NullTrigger`, `Event`, `First`, `Combine`, `Lock`, `with_timeout`/`SimTimeoutError`, `cocotb.queue.*`, `cocotb.types.{Logic,LogicArray,Range}` (X/Z→0), task handles with `kill()`/`join()` and full trigger deregistration on cancellation, real `cocotb.*` facade submodules, DUT `_name`/`_log`/`__dir__`/case-insensitive lookup, `len(signal)`, width-masked writes, width-correct signed conversion.

### Conformance evidence
- `tests/cocotb_shim/test_shim_conformance.py` — 17 tests covering every Timing / Tasks-and-Triggers / Handles item in the spec, against a real generated model.
- `tests/cocotb_shim/test_axil_cocotbext.py` — **unmodified installed cocotbext-axi** AXI-Lite master vs `AxilRegs.arch`: register R/W, byte strobes, queued transactions, 50-op random soak. 4/4 pass.
- `tests/cocotb_shim/test_axi4_cocotbext.py` — cocotbext-axi AXI4 master vs `Axi4Mem.arch`: INCR bursts, byte strobes, independent AW/W/B/AR/R backpressure via pause generators, queued transactions, final byte-addressed memory comparison. 5/5 pass.
- Full suite: **903 passed, 0 failed** (integration tests skip gracefully when pybind11/cocotbext-axi are absent).
- Legacy compatibility: `python/test_arch_cocotb_smoke.py` passes unchanged; existing cvdp tests use explicit `units=` so the cocotb-standard `Timer` default (`step`) breaks nothing. The `test_mshr_cocotb.py` failures are pre-existing environmental (`cocotb-config`/iverilog flow), reproduced identically on `main`.

### Docs
`doc/arch_sim_cocotb.md` rewritten for the new time/scheduler model and bus-library usage; CLAUDE.md pybind section updated.

### Known limitation (pre-existing)
The pybind layer truncates ports wider than 64 bits (`emit_wide_binding` caps at 2 words), so pybind-driven bus data widths should stay ≤ 64 bits. Noted in the docs; fixing the wide binding is a separate change.


teabone113:
Opened in error — retracting.

PR #741, opened 2026-07-31
fix(memory): emit inference-safe RAM and FIFO structures

teabone113:
## Summary

Make the generic SystemVerilog emitted for RAMs and FIFOs follow portable FPGA memory-inference structures:

- add explicit FIFO `latency 1` support for synchronous and dual-clock FIFOs, retaining the existing ready/valid FWFT interface with a registered head-data path
- keep payload memories out of reset branches and emit reset-free synchronous read/write processes
- emit true-dual-port RAM ports in separate clocked processes and honor each port's own clock
- retain `latency 0` as the existing combinational-read, LUTRAM-friendly option
- mirror the new behavior in native simulation and validate unsupported latency/clock combinations
- document the timing choice and update the bundled ARCH skill snapshots

The generated RTL remains vendor-neutral: this does not instantiate Xilinx XPM/RAMB or another vendor primitive.

## Root cause

The previous FIFO emitter only had a combinational memory-read shape. Deep FIFOs therefore carried a memory mux on `pop_data`, and memory access was mixed with reset/control logic in ways that prevent reliable block-RAM inference.

The true-dual RAM path also did not model the two physical ports independently. In particular, port B could be emitted under port A's clock, and a combined multi-write process is not a recognized true-dual-port inference structure in Vivado.

## Impact

Designers can now choose:

- `latency 0` for a combinational-read FIFO
- `latency 1` for a registered FWFT FIFO whose memory processes are suitable for FPGA block-RAM inference

True-dual RAMs preserve independent port clocks in both generated SV and native simulation. Existing latency-zero behavior remains explicit and unchanged.

## Validation

- `cargo fmt --check`
- `scripts/sync_skill_snapshots.sh check`
- complete `cargo test` suite: 908 passed
- focused regressions cover synchronous and asynchronous latency-1 FIFOs, native-simulator behavior, latency validation, single-RAM latency-2 driver ownership, independent true-dual clocks, invalid clock counts, and two-clock native simulation

## Maintainer note

The branch includes a `0.71.1` package-version bump used for downstream validation. That commit can be dropped if upstream release versioning should remain separate from the functional change.


teabone113:

teabone113:

teabone113:
CLA workflow note: I posted the exact signing phrase in a new comment, and CLA Assistant recognized it—the bot comment now says all contributors have signed. However, the workflow failed while persisting the signature:

> Could not update the JSON file: Could not update file: 7 of 7 required status checks are expected.

Run: https://github.com/arch-hdl-lang/arch-com/actions/runs/30606900336

`signatures/cla.json` on `main` still does not contain `teabone113`, so a `recheck` on the other PRs will currently fail as well. It appears branch protection is preventing CLA Assistant from committing the signature; a maintainer will likely need to allow the action/token to bypass that protection or record the signature manually before rechecking the PRs.

nogate-ai (changes requested):
Requesting one scoped change to the true-dual RAM clock mapping. Please keep the separate per-port processes and independent-clock support; that RTL structure is correct for reliable Xilinx BRAM inference.

Remove the declaration-order fallback and enforce an unambiguous ARCH mapping:

- With one clock, use it as the shared clock for both RAM ports.
- With two clocks, each port group g must have clk_g. For the specified a/b groups, that means a uses clk_a and b uses clk_b.
- Missing, partial, or mixed two-clock naming should be a type-check error rather than a positional fallback.

Reason: Vivado BRAM inference recognizes the memory and clocked-process structure, not the spelling of signal identifiers. The clka/clkb identifiers in AMD examples are illustrative, not required names. See the official memory inference capabilities and true-dual-port example:

https://docs.amd.com/r/en-US/ug901-vivado-synthesis/Memory-Inference-Capabilities
https://docs.amd.com/r/en-US/ug901-vivado-synthesis/Dual-Port-Block-RAM-with-Two-Write-Ports-in-Read-First-Mode-Verilog-Example

The positional fallback adds no Xilinx inference benefit and makes the hardware meaning depend on declaration order. In the current implementation, declaring clk_b plus other_clk causes both port processes to select clk_b, silently ignoring other_clk.

Please add coverage for the exact two-clock mapping, the one-shared-clock form, and rejection of partial/mixed naming. The other RAM/FIFO inference changes should remain.

nogate-ai (approved):
Maintainer update applied; no action is required from the author. I kept the approved RAM/FIFO inference changes and preserved the two original commits and their authorship. The positional two-clock fallback was removed: a one-clock RAM still accepts any shared clock name, while a two-clock true-dual RAM requires clk_<port-group> (for example clk_a and clk_b). This makes clock selection explicit and stable, while preserving separate per-port always_ff processes needed for Xilinx BRAM inference. I also aligned native latency-2 RAM simulation with SystemVerilog nonblocking-assignment timing and added regression coverage. The independent-clock tests, FIFO latency-1 tests, full 851-test integration suite, and 130 library tests pass.

PR #742, opened 2026-07-31
fix(elaborate): rewrite derived-parameter variants consistently

teabone113:
## Summary

Fix elaboration of parameterized child modules whose defaults contain derived parameters.

Variant discovery and instance-name rewriting now:

- use the same evaluated raw override map
- merge overrides with the child module's defaults
- recompute derived child parameters in declaration order
- keep explicitly overridden derived parameters pinned

## Root cause

Variant discovery already recomputed defaults such as:

```arch
param ROW_MODE: const = 0;
param MODE_WIDTH: const = ROW_MODE + 1;
```

Instance-name rewriting did not perform the same recomputation. If sibling wrappers instantiated the child with different `ROW_MODE` values, discovery emitted suffixed variants with the correct `MODE_WIDTH`, but rewriting searched with a stale derived value. It found no match and left an unsuffixed module name that was no longer emitted.

## Impact

Parameterized hierarchies now produce self-consistent module definitions and instance references when a base parameter changes a dependent default. The change is confined to elaboration; it does not alter parameter behavior for modules with a single effective variant.

## Validation

- `cargo fmt --check`
- focused `test_sibling_wrappers_rewrite_child_variants_with_derived_params`
- complete `cargo test` suite: 901 passed

The regression builds two sibling wrappers with different base-parameter overrides and verifies that each instance selects the fully evaluated child specialization.


teabone113:

nogate-ai (approved):
Reviewed the derived-parameter variant rewrite and the current-main port. Variant discovery and instance rewriting now share override evaluation and derived-parameter recomputation; the regression and full local suite pass.

PR #743, opened 2026-07-31
feat(sim): expand and embed native cocotb compatibility

teabone113:
## Summary

Expand `arch sim --pybind --test` into a substantially more complete cocotb-compatible native simulation flow, and make an installed `arch` binary self-contained.

Highlights:

- event-driven scheduler with exact integer-picosecond time
- `RisingEdge`, `FallingEdge`, `Edge`, `Timer`, `ClockCycles`, `Clock`, `ReadOnly`, `Event`, `First`, and timeouts
- cocotb-style task handles, cancellation cleanup, queues, and two-state `Logic`/`LogicArray`
- width-aware signal handles, arbitrary-width Python integer transfer, DUT discovery, and uninitialized-input markers
- unmodified `cocotbext-axi` 0.1.28 AXI4 and AXI4-Lite masters against native ARCH models
- native-simulator parity fixes for parameter specializations, named structs, bus wires, falling-edge seq blocks, FIFO payloads, pipelines, and wide values
- embedded `arch_cocotb` and `cocotb` shim packages, materialized beneath the simulation output directory so copied/installed binaries do not need the source tree or `ARCH_PYTHON_DIR`
- native and real-cocotb/Verilator conformance scripts

The implementation remains intentionally two-state and does not attempt to reproduce the complete GPI/VPI phase model. Those portability differences are documented in `doc/arch_sim_cocotb.md`.

## Root cause

The original shim exposed only a small trigger/task surface and depended on locating Python packages beside the source checkout. This limited ordinary cocotb test reuse, prevented broader bus-library compatibility, and made a copied compiler binary fail unless callers manually set `ARCH_PYTHON_DIR`.

Several native code-generation gaps were also exposed once realistic cocotb and `cocotbext-axi` tests exercised parameterized, wide, struct-valued, FIFO, FSM, and pipeline designs.

## Impact

Users can ship a single `arch` executable and run cocotb-style tests directly against generated native models. Existing conventional cocotb tests can also be cross-checked against emitted SystemVerilog through Verilator using the included script.

This branch is based directly on `main`; it does not include the RAM/FIFO inference or derived-parameter fixes from the other submissions.

## Validation

- `cargo fmt --check`
- complete `cargo test` suite: 916 passed
- `scripts/sync_skill_snapshots.sh check`
- `scripts/test_native_cocotb.sh`: 10 scheduler/API tests plus AXI4-Lite, AXI4, and 130-bit native pybind cases
- `scripts/test_verilator_cocotb.sh`: the same three designs passed with cocotb 2.0.1, cocotbext-axi 0.1.28, and Verilator 5.050
- embedded-runtime regression copies the binary and runs without `ARCH_PYTHON_DIR`

## Maintainer note

The branch includes a `0.72.1` package-version bump used for downstream validation. That version change can be dropped if upstream release versioning should be handled separately.


teabone113:

nogate-ai (approved):
Maintainer review complete. I ported the original commits onto current main while preserving their author metadata, adapted the implementation to the split elaborate/sim-codegen layout, dropped the unrelated 0.72.1 version bump, and fixed the pipeline concat width regression so plain arithmetic uses its widened ARCH result width (for example, UInt<8> + UInt<8> occupies 9 bits). Added a regression for that case. Full cargo test, native cocotb AXI/AXI-Lite/wide tests, Verilator cocotb cross-checks, formatting, skill snapshots, and release-doc checks pass locally. No author action is required.

PR #744, opened 2026-07-31
fix(pipeline): preserve per-register init and reset semantics

### LINKED ISSUE #740: pipeline codegen conflates `init` and `reset`, resets `reset none` stage registers
## Summary

The first-class `pipeline` emitters conflate three distinct register properties:

1. SystemVerilog declaration initialization (`init`)
2. Clocked reset behavior (`reset SIGNAL => VALUE`)
3. No reset (`reset none`)

The SV emitter derives one `init_str` using **reset value → explicit init → otherwise zero**, then uses that value both on the signal declaration and in the pipeline-wide reset branch. This adds declaration initializers that were not requested, resets `reset none` registers, and uses the reset value instead of an explicitly different `init` value.

The native pipeline simulator has the corresponding semantic problem: it constructs every stage register at zero and resets every stage register through one pipeline-wide `_do_reset()`.

Confirmed on current upstream `main` at `06ffffa` and independently observed with `arch 0.72.1` / Verilator 5.050.

## Reproduction

```arch
pipeline PipelineInitReset
  port rst: in Reset<Sync, High>;
  port a: in UInt<8>;
  port b: in UInt<8>;
  port c: in UInt<8>;
  port d: in UInt<8>;

  stage StageOne
    reg reset_only: UInt<8> reset rst => 3;
    reg reset_none: UInt<8> reset none;
    reg init_only: UInt<8> init 5 reset none;
    reg init_and_reset: UInt<8> init 7 reset rst => 9;

    seq on clk rising
      reset_only <= a;
      reset_none <= b;
      init_only <= c;
      init_and_reset <= d;
    end seq
  end stage StageOne
end pipeline PipelineInitReset
```

Current SV declarations:

```systemverilog
logic [7:0] stageone_reset_only = 3;
logic [7:0] stageone_reset_none = 0;
logic [7:0] stageone_init_only = 5;
logic [7:0] stageone_init_and_reset = 9;
```

Current reset branch:

```systemverilog
if (rst) begin
  stageone_reset_only <= 3;
  stageone_reset_none <= 0;
  stageone_init_only <= 5;
  stageone_init_and_reset <= 9;
end
```

The correct behavior is:

| ARCH declaration | Declaration initializer | Reset assignment |
|---|---|---|
| `reset rst => 3` | none | `3` |
| `reset none` | none | none |
| `init 5 reset none` | `5` | none |
| `init 7 reset rst => 9` | `7` | `9` |

## Additional reset-selection defect

The SV pipeline emitter calls `extract_reset_info(&p.ports)` once and puts all stage data registers in that one reset process. A stage register naming a second reset signal or a different polarity is therefore reset by the first reset port instead of its declared reset.

For example, with `rst_a: Reset<Sync, High>` followed by `rst_b_n: Reset<Sync, Low>`, both of these:

```arch
reg reset_a: UInt<8> reset rst_a => 1;
reg reset_b: UInt<8> reset rst_b_n => 2;
```

are currently emitted under `if (rst_a)`.

## Native simulation

`src/sim_codegen/pipeline.rs` also loses the distinction:

- Constructor initializes every stage register to zero, ignoring explicit `init`.
- `_do_reset()` writes every stage register, including `reset none`.
- Reset triggering is selected from one pipeline-wide reset port.

For the reproduction above, the generated constructor starts all four registers at zero. `_do_reset()` writes `3, 0, 0, 9`, so `init_only` is ignored even though the SV declaration currently initializes it to 5. Fixing only SV codegen would leave `arch sim` and generated SV inconsistent.

## Root cause

In `src/codegen/pipeline.rs`, stage registers are represented as `(signal_name, type_str, init_str)`. The `init_str` is constructed as:

```rust
if let Some(reset_val) = reset_value_expr(&r.reset) {
    reset_val
} else if let Some(init_expr) = r.init {
    init_expr
} else {
    0
}
```

That same string drives declaration initialization and reset assignments, so the representation cannot express the documented semantics.

## Impact

- Verilator 5.050 reports `PROCASSINIT` for reset-only pipeline registers because they have both declaration initialization and procedural assignment.
- Unrequested target-dependent power-up semantics are introduced.
- `reset none` registers gain reset behavior and associated hardware/control semantics.
- Explicitly different `init` and reset values are miscompiled.
- Multiple reset signals/polarities are ignored for stage data registers.
- Native simulation does not match emitted SV initialization.

This is a correctness bug rather than only a lint-quality issue.

## Suggested fix direction

Preserve separate stage-register metadata:

- `decl_init: Option<Expr>` from `r.init` only
- full `RegReset` metadata (signal, sync/async, polarity, value)
- register vs. combinational-wire classification

Then:

1. Emit a declaration initializer only from explicit `init`.
2. Reset only registers whose `RegReset` is not `None`.
3. Partition stage-register updates by reset specification, following the ordinary module emitter's approach. This matters for async-reset processes: reset-free registers cannot safely be placed in a reset-sensitive process.
4. Keep pipeline-generated `valid_r` and wait-state FSM registers on the pipeline control reset.
5. Make native-model constructor values follow `init`, while reset handling touches only reset-bearing registers.
6. Decide explicitly how `flush ... clear` handles stage registers with `reset none`: skip them or reject that combination, since they have no declared reset value.

## Regression coverage

- Reset-only stage register: no declaration initializer; reset assignment retained.
- `reset none`: no declaration initializer and no reset assignment.
- Init-only register: declaration initializer retained and reset does not change it.
- Different `init` and reset values: both values emitted in their correct roles.
- Multiple reset signals, polarities, and sync/async kinds.
- Mixed reset-bearing and reset-free assignments in one stage `seq` block.
- Native model initialization/reset parity with generated SV.
- Verilator 5.050 `--lint-only --Wall` has no `PROCASSINIT` for reset-only pipeline registers.

The existing `simple_pipeline` SV snapshot currently includes `= 0` for reset-only stage registers and will need updating.

teabone113:
Closes #740.

## Summary

Preserve the documented per-register distinction between declaration initialization, clocked reset behavior, and `reset none` in first-class pipelines.

The SystemVerilog and native-simulation emitters now:

- emit declaration initializers only for explicit `init`
- reset only registers with a declared reset
- preserve different `init` and reset values
- group clocked updates by each register's actual reset signal, polarity, and sync/async kind
- keep reset-free registers out of reset-sensitive processes
- initialize native model storage from `init` and apply only declared resets
- retain pipeline control-state reset handling separately from stage data-register reset handling

## Root cause

Pipeline stage registers were reduced to a single `init_str`, chosen from reset value, explicit init, or zero. That one value was then reused for both the SV declaration and the pipeline-wide reset branch. The representation could not distinguish:

- `init 7 reset rst => 9`
- `reset rst => 9`
- `init 7 reset none`
- `reset none`

The native emitter had the corresponding pipeline-wide constructor/reset behavior. Both backends also selected one reset port for every stage register, even when individual declarations named different resets.

## Impact

This removes unintended power-up state, stops resetting `reset none` registers, eliminates the generated `PROCASSINIT` pattern for reset-only registers, and restores native/SV parity. Mixed reset signals and polarities now retain their source-level semantics.

## Validation

- `cargo fmt --check`
- focused `test_pipeline_preserves_per_register_init_and_reset_semantics`
- complete `cargo test` suite: 901 passed
- updated pipeline snapshots cover the changed declaration and process structure

The focused fixture covers reset-only, reset-none, init-only, different init/reset values, multiple reset signals, active-high/active-low behavior, and native-model initialization/reset handling.


teabone113:

PR #745, opened 2026-07-31
fix(codegen): scope package imports to source files

teabone113:
## Summary

Keep package imports with the ARCH source file/design units that own the corresponding `use` declarations.

The code generator now:

- records the byte range of each original source in a combined multi-file parse
- resolves the active, deduplicated `use` set for each emitted module or FSM from its source span
- keeps unrelated generated SV files free of leaked imports
- lets elaborated parameter specializations inherit the imports of their originating source item
- preserves existing single-source in-memory codegen behavior

## Root cause

Module and FSM emission scanned every `Item::Use` in the combined AST. In a multi-file build, that made `use` effectively compilation-global. Repeated uses were emitted repeatedly, and unrelated design units received wildcard imports from other source files.

Because wildcard imports are emitted at compilation-unit scope, a large multi-file hierarchy multiplied this into extensive Verilator `IMPORTSTAR` diagnostics and exposed unrelated units to package namespace collisions.

## Impact

For a build containing two package consumers and an unrelated module:

- each consumer now receives one import
- duplicate `use` declarations are deduplicated
- the unrelated module receives no import
- generated parameter variants retain only their source file's imports

Downstream report and reproduction: [teabone113/arch-com#3](https://github.com/teabone113/arch-com/issues/3).

## Validation

- `cargo fmt --check`
- focused `test_duplicate_use_emits_one_package_import_per_module`
- focused `test_multifile_package_imports_stay_with_owning_source_file`
- complete `cargo test` suite: 902 passed

The multi-file regression covers duplicate uses, two consumers, an unrelated module, and parameter-specialized output ownership.


teabone113:

Issue #750, opened 2026-08-01
Enhancement: report all accumulated diagnostics, not just the first — arch check discards N-1 of N errors


nogate-ai:
## Summary

`arch check` reports **exactly one diagnostic per invocation**, no matter how many independent errors the file has. The type checker already finds and accumulates all of them — `main.rs` throws the rest away at ten separate call sites. Fixing this is mostly plumbing: batch the accumulated `Vec<CompileError>` into one `miette` report via `#[related]`.

## Current state (verified on a freshly built `origin/main`, 06ffffa)

`src/typecheck.rs:105` accumulates:

```rust
pub errors: Vec<CompileError>,
```

and `check()` (`src/typecheck.rs:142`) returns **all** of them:

```rust
) -> Result<(Vec<CompileWarning>, HashMap<usize, usize>), Vec<CompileError>> {
    ...
    if self.errors.is_empty() { Ok(...) } else { Err(self.errors) }
}
```

The error sites push and continue — they do not bail. E.g. the width-mismatch arm at `typecheck.rs:2956` pushes and falls through.

Then `src/main.rs` discards everything but element 0, at **ten** sites (3656, 3668, 3674, 3682, 3689, 3739, 3771, 3778, 3784, 3791):

```rust
let err = errs.into_iter().next().unwrap();
return Err(ms.report_error(err));
```

`report_error` (`main.rs:511`) is `fn(CompileError) -> Report` — single-error by signature.

## Proof that errors are being discarded, not just absent

```arch
module Two
  port a: in UInt<8>;
  port b: in UInt<8>;
  port o1: out UInt<8>;
  port o2: out UInt<8>;
  comb
    o1 = a + b;
    o2 = a + b;
  end comb
end module Two
```

Two identical, independent width errors. Run 1 reports only `o1` (line 7). Fix `o1` only; run 2 reports `o2` (line 8). Both were in the `Vec` the whole time.

## The second half: errors surface in *pass* order, not source order

Worse than "one at a time" — the *sequence* is unpredictable, because whichever pass fails first wins and later passes never run. Starting from a file with four independent errors:

```arch
o1 = a + b;              // width mismatch
o2 = missing_one;        // undefined signal
o3 = missing_two;        // undefined signal
o4 = a.trunc<99>();      // widening trunc
```

[...]

PR #755, opened 2026-08-01
fix(comb-loop): give each top-level module its own graph namespace

nogate-ai:
## Summary

`analyze_whole_design` expanded **every** top-level module into one shared `GraphBuilder` at the same root instance path (`vec![]`), so all top-level modules shared a single `(inst_path, signal)` node namespace — `ModA`'s `p` and `ModB`'s `p` interned to the same node.

Distinct top-level modules are by definition unconnected, so no combinational path can run between them. The shared namespace could therefore only do harm, in two ways.

### 1. False positives (the reason this matters)

Two individually-acyclic top-level modules that happen to use the same signal names get their chains stitched into a cycle that exists in neither:

```arch
module ModA          module ModB
  ...                  ...
  q = i;               p = i;
  p = q +% 1;          q = p +% 1;
  o = p;               o = q;
```

`ModA` is `q → p`. `ModB` is `p → q`. Each is a straight line; neither has a loop.

```
$ arch check ModA.arch          → OK: no errors
$ arch check ModB.arch          → OK: no errors
$ arch check ModA.arch ModB.arch
warning: whole-design combinational feedback cycle (2 nodes) involving modules [ModB]; cycle: p -> q -> p
```

Since the repo convention is one construct per file, multi-file invocation is the normal way to check a design — and common HDL signal names (`valid`, `ready`, `cnt`, `state`) make the collision easy to hit.

### 2. Misattribution

`path_owner` is keyed on the instance path, so the entry for `vec![]` was overwritten by whichever top was expanded last. A cycle in one module got reported as `involving modules [SomeOtherModule]`, naming an **acyclic** module and pointing the warning span at that module's declaration. Note in the repro above that the fabricated cycle is blamed on `ModB` — arbitrarily.

## Fix

Each top-level module now gets its own `GraphBuilder` and its own Tarjan run; the per-SCC classification is unchanged and results aggregate across tops.

Node display and warning text are untouched, so single-top designs — the common case, and what every existing test covers — produce byte-identical output.

Also switches top-level discovery from iterating `module_by_name` (a `HashMap`, hence nondeterministic order) to source declaration order. That nondeterminism meant *which* module got blamed could vary between runs of the same input.

## Trade-off

`per_output_cache` memoizes each child module's comb-dep map and now lives per-top rather than being shared across all tops, so a module instantiated under two different top-level modules has its body walked once per top instead of once total. This is a body walk over already-parsed AST during `arch check`, bounded by (number of tops × distinct child modules); it did not move the wall clock on the full suite. Correct attribution is worth more than the shared cache — and a cache shared across tops is exactly what conflated their namespaces in the first place.

## Testing

- `test_sibling_top_modules_sharing_signal_names_is_not_a_cycle` — the false-positive repro above.
- `test_comb_loop_attributed_to_the_owning_top_module` — cycle in the second-declared module must name that module, not the acyclic first one.

Both were **verified to fail without the src change** (reverted `comb_graph.rs`, kept the tests, confirmed both fail) so they genuinely pin the behavior.

Full suite green: 750/750 `integration_test` (including all 13 pre-existing comb-loop tests), 68 lib, 23 fp, 14 param-where, 14+3+7 pipelined-ops. `cargo fmt --check` clean.

`formal_test` has 2 failing Lean tests (`unknown module prefix 'ArchConstructProof'`) — pre-existing and environmental, confirmed identical on unmodified `origin/main` in this worktree; the Lean `.lake/build` dir doesn't exist here.

## Scope

Internal analysis fix — no user-facing syntax or semantics change, no grammar surface touched (`comb_graph.rs` only), so no doc update applies. The comb-loop check remains a **warning**; this PR does not change its severity.

Found while investigating whether the comb-loop detector needed attention as a follow-up to the diagnostics work in #750/#748.

### DIFF
--- a/src/comb_graph.rs
+++ b/src/comb_graph.rs
@@ -1198,7 +1198,10 @@ pub struct WholeDesignAnalysis {
 /// 1. Identify top-level modules (modules not instantiated anywhere).
 /// 2. For each top-level, recursively flatten the instance hierarchy into
 ///    a directed node-and-edge set, where each node is `(inst_path, signal)`.
-/// 3. Run Tarjan's SCC.
+///    Each top gets its own graph — tops are mutually unconnected, so a
+///    namespace shared across them can only fabricate cycles out of
+///    same-named signals in unrelated modules.
+/// 3. Run Tarjan's SCC per top.
 /// 4. Tag each non-trivial SCC with the set of owning-parent paths;
 ///    classify as suppressed if any owning module has the pragma.
 pub fn analyze_whole_design(source: &SourceFile, symbols: &SymbolTable) -> WholeDesignAnalysis {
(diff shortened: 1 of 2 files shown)


nogate-ai:
## Long verification: differential sweep over the full `.arch` corpus

Ran `arch check` over all **868** `.arch` files in the repo with this branch's binary and with a freshly built `origin/main` binary, diffing every comb-loop line.

- **161 comb-loop warning lines on each side.**
- Exactly one line differed — and it turned out to be **pre-existing run-to-run noise, not this change**.

The one differing line (`tests/e203/e203_ifu.arch`) reports the same cycle both ways: same 8 nodes (verified as identical *sets*), same `[e203_ifu, e203_ifu_ifetch, e203_ifu_ift2icb]` attribution, same span. Only the rotation the path is printed in differs.

Running the **unmodified `main`** binary three times on that one file gives two different orderings on its own:

```
run 1: ifetch.ifu_rsp_ready -> ... -> icb.stall_pipe -> ... -> icb.ifu_req_ready -> ifetch.ifu_rsp_ready
run 2: icb.stall_pipe -> ... -> ifetch.ifu_rsp_ready -> ... -> icb.ifu_req_ready -> icb.stall_pipe
run 3: (same as run 2)
```

So detection, module attribution, and node sets are **unchanged across the entire corpus**; the only variation is a rendering order that was already unstable before this PR.

## Correction to the PR description

The description says this PR switches top-level discovery to source order because the `HashMap` iteration "meant *which* module got blamed could vary between runs." That is accurate for the source it removes, but I should be explicit that **it does not make the output deterministic overall** — I measured 8 runs on `e203_ifu.arch` after this change and still get 4 distinct orderings, the same count as before.

The residual nondeterminism is in the cycle *path rendering* only. Across those 8 runs the node count (`8 nodes`) and the module list are stable — as they must be, since the SCC partition is a graph invariant and only the traversal order within an SCC varies. It comes from `HashSet`/`HashMap` iteration further down in `expand_module` (e.g. `bus_values`, the `per_output_comb_deps` maps) affecting node interning order.

That is pre-existing, cosmetic, and out of scope here, but it makes warning text unstable for golden-output tests, so I've filed it separately rather than leaving it buried in this thread.

## Suite status

750/750 `integration_test` (incl. all 13 pre-existing comb-loop tests), 68 lib, 23 fp, 14 param-where, 14+3+7 pipelined-ops. `cargo fmt --check` clean. The 2 `formal_test` Lean failures are environmental and reproduce identically on unmodified `main` in this worktree.

Issue #748, opened 2026-08-01
arch check accepts undefined value identifiers — emits SV with undeclared nets, reports "OK: no errors"


nogate-ai:
## Summary

`arch check` accepts references to identifiers that are **not declared anywhere** — no port, no `wire`, no `reg`, no `let`, no `param`. It prints `OK: no errors` and exits 0. `arch build` then emits SystemVerilog referencing undeclared nets, which Verilator rejects.

The `UndefinedName` diagnostic and its `CompileError::undefined(..)` constructor already exist; they are simply never reached for value-expression identifiers.

## Repro (verified on a freshly built `origin/main`, 06ffffa)

```arch
module Undef
  port a: in UInt<8>;
  port o2: out UInt<8>;
  port o3: out UInt<8>;

  comb
    o2 = missing_one;
    o3 = missing_two;
  end comb
end module Undef
```

```
$ arch check Undef.arch
OK: no errors           # exit 0

$ arch build Undef.arch -o undef.sv
Wrote undef.sv

$ verilator --lint-only undef.sv
%Error: undef.sv:7:15: Can't find definition of variable: 'missing_one'
%Error: undef.sv:8:15: Can't find definition of variable: 'missing_two'
```

Emitted SV:

```sv
assign o2 = missing_one;
assign o3 = missing_two;
```

## Scope — every value-expression context is affected

All four of these pass `arch check` clean:

| Context | Snippet |
|---|---|
| `let` initializer | `let x: UInt<8> = ghost_let;` |
| `seq` RHS | `r <= ghost_seq;` |
| arithmetic sub-expression | `o = a + ghost_expr;` |
| `if` condition in `comb` | `if ghost_cond ... end if` |

## Root cause

`src/typecheck.rs:4034` — the `ExprKind::Ident` arm of `resolve_expr_type`:

```rust
ExprKind::Ident(name) => {
    if let Some(ty) = local_types.get(name) {
        ty.clone()
    } else {
        // Check param (treat as generic width)
        Ty::Error
    }
}
```

An unresolved name falls through to `Ty::Error`, which is the poison value that suppresses downstream diagnostics — so not only is no error raised here, the surrounding expression's checks go quiet too. No diagnostic is ever pushed.

[...]

Issue #737, opened 2026-07-31
inst boundary: Clock/Reset port connections are unchecked — Reset<Async,Low> into a Reset<Sync,High> port ships inverted reset, silent in both arch check and Verilator


nogate-ai:
## Summary

`inst` port connections perform **no type conformance check on `Clock` and `Reset` ports**. A `Reset<Async, Low>` signal can be connected to a `Reset<Sync, High>` port, and a plain `Bool` can be connected to either a `Clock<D>` or a `Reset<K,P>` port. All are accepted silently.

The reset case produces **functionally inverted hardware that no tool in the flow detects**: `arch check` is silent, and `verilator --lint-only` is also silent, because after lowering both sides are a 1-bit `logic`. The polarity/kind information exists only in ARCH's type system, and ARCH discards it at the instance boundary.

Verified on a freshly built `origin/main` binary (arch 0.70.8, `a8c930e`) with Verilator 5.048.

## Reproduction

`Child.arch` — declares a **sync, active-high** reset:

```arch
module Child
  port clk: in Clock<AlphaDom>;
  port rst: in Reset<Sync, High>;
  port d:   in UInt<8>;
  port reg q: out UInt<8>;
  seq on clk rising
    if rst
      q <= 0;
    else
      q <= d;
    end if
  end seq
end module Child
```

`Parent.arch` — drives it with an **async, active-low** reset, and with a clock from a different domain:

```arch
module Parent
  port clk_a: in Clock<AlphaDom>;
  port clk_b: in Clock<BetaDom>;
  port rst_n: in Reset<Async, Low>;
  port din: in UInt<8>;
  port dout: out UInt<8>;

  inst c: Child
    clk <- clk_b;
    rst <- rst_n;
    d <- din;
    q -> dout;
  end inst c
end module Parent
```

```
$ arch check Parent.arch Child.arch
OK: no errors

$ arch build Parent.arch Child.arch && verilator --lint-only Parent.sv Child.sv
(clean — exit 0)
```

Emitted SV:

```sv
// Parent.sv
Child c (
  .clk(clk_b),
  .rst(rst_n),     // Async, active-LOW
  ...
);

// Child.sv
always_ff @(posedge clk) begin
  if (rst) begin   // treated as Sync, active-HIGH
```

So the child is **held in reset during normal operation and released during reset**, and the async release is lowered as synchronous (no `posedge rst` in the sensitivity list). Both failures ship silently.

## Full matrix (all pass `arch check`)

[...]

PR #793, opened 2026-08-06
fix(tools): resolve sibling SV lint deps by dependency closure, not directory glob

nogate-ai:
## Problem

`tools/run_arch_regression.py`'s `sibling_sv_deps()` globbed **every** `*.sv` in a unit's directory as a Verilator lint dependency, excluding only the stems the unit itself generates.

That is safe when a directory keeps a handful of legacy submodules as checked-in SV. It is wrong wherever a test tree also checks in the *generated* `.sv` next to each `.arch` — `arch build` inlines transitive submodules, so those siblings redeclare each other's modules.

`tests/e203/` is the worst case: 62 checked-in `.sv`, **11 module names declared in more than one file** (`e203_exu.sv` alone carries 10 modules, `e203_ifu_ifetch.sv` carries 4). Whenever the whole-directory `arch check` group fails — which it does today, see #TBD — `discover_units()` falls back to per-file units, and each unit then drags in all 61 siblings. Verilator dies on MODDUP before it ever looks at the unit under test:

```
%Warning-MODDUP: .../e203_exu_alu.sv:5:8: Duplicate declaration of module: 'e203_exu_alu'
                 .../e203_exu.sv:573:8: ... Location of original declaration
```

So `python3 tools/run_arch_regression.py --pattern "e203/*"` reported spurious `fail at verilator_lint` for nearly every e203 unit, independent of any actual defect in those units.

## Fix

Compute the dependency closure from the generated SV instead of globbing: start with the module types it instantiates but does not declare, pull in the siblings that declare them, and repeat until the reference set is satisfied. A sibling that would redeclare an already-defined module is skipped.

## Result

| | before | after |
|---|---|---|
| pass | 1 | **53** |
| fail | 57 | **5** |
| skip_check_failed | 4 | 4 |

(`--pattern "e203/*" --skip-sim`)

The 5 remaining failures are **real signal that the MODDUP flood was masking** — two `arch build` codegen defects and one fixture drift. They are reported separately rather than fixed here, so this PR stays a single-purpose tooling fix.

## Verification

- **No behavior change on the directories that genuinely depend on orphan sibling `.sv`** (the case the original glob existed to serve): `aes`, `assert_cover`, `fp_v1` give 7 pass / 6 skip / 1 fail both before and after. The `fp_v1` failure is pre-existing on `main`.
- `cargo test --release`: unaffected — this PR touches no Rust. Two `formal_test` Lean-replay tests fail in this worktree because `proofs/lean_thread_lowering/.lake/build` has never been built here (`unknown module prefix 'ArchConstructProof'`); that is environmental and reproduces without this change.


nogate-ai:
## Long verification complete — full `tests/` tree, before vs after

Ran the whole tree both ways (`--release --skip-sim --jobs 4`), baseline from `origin/main`:

| | baseline | with fix |
|---|---|---|
| pass | 266 | **561** |
| fail | 316 | **21** |
| skip_check_failed | 73 | 73 |

- **Regressions (pass → not pass): 0**
- Newly passing: 295

`tests/e203/` was the case that prompted this, but `tests/cvdp/` was hit harder in absolute terms — 273 fixtures in one directory, each dragging in the other 272 as lint deps.

All 21 remaining failures were also failing on baseline; they were simply drowned in the MODDUP noise. They break down as:

- 10 `cvdp__*`, 3 `formal__*`, `fp_v1`, `verilog_eval__Prob084`, `regression__issues__fsm_param_vec_scalar_widths` — pre-existing, unrelated to this change.
- 5 `e203__*` — the real signal this PR unmasked:
  - `e203_dtcm_ram`, `e203_itcm_ram`, `e203_srams`: `arch build` hoists a `for`-loop variable into an auto-emitted bounds assertion at module scope, so the emitted SV does not elaborate (`Can't find definition of variable: 'i'`).
  - `e203_biu`: emitted SV references `sel_ppi_r`/`sel_clint_r`/`sel_plic_r`/`sel_fio_r`/`sel_mem_r`, which are used but never declared in `e203_biu.arch` — a real-design reproducer for #748 (undefined value identifiers pass `arch check` clean).
  - `e203_exu`: `PINNOTFOUND` on `o_alu`/`o_agu`/… — fixture drift plus #737 (`inst` connections are untyped, so naming a port the leaf does not have is silently accepted).

Filed separately; none are fixed here.

nogate-ai:
## Follow-up for a maintainer decision: the nightly baseline is stale because of this bug

`tests/arch_regression_baseline.json` lists **187** units, and `scripts/nightly_equivalence.py` passes it as `--baseline`, which restricts the nightly sweep to exactly those names. That baseline was generated from a run made under the bug this PR fixes, so it encodes the broken result — it contains a single e203 entry (`e203__e203_itcm_icb_bus`), the one unit that happened to survive the MODDUP flood.

With this PR, 561 units pass. So ~370 currently-passing units are excluded from the nightly gate for no reason other than the bug.

I have **not** refreshed the baseline here. It would roughly triple what nightly gates on, which changes the CI contract and could pull in units that are green locally but flaky in CI — a maintainer call, not a drive-by. When you want it:

```bash
python3 tools/run_arch_regression.py --release --update-baseline tests/arch_regression_baseline.json --allow-failures
```

Worth doing as its own PR so the coverage expansion is reviewable separately from the fix.

(#794, stacked on this branch, adds one entry by hand — `e203__e203_ifu_ifetch` — so its rewritten testbench actually runs.)

Issue #815, opened 2026-08-08
arch build drops a ram's `ports` groups when emitting its .archi, making separate compilation of ram consumers impossible


nogate-ai:
## Summary

`arch build` does not round-trip a `ram`'s `ports` groups into the generated `.archi`. Any consumer that resolves the ram through separate compilation then fails with `ram \`X\` has no port groups`, and there is no way to write a working `.archi` ram stub.

## Reproduction

Verified on a freshly built binary.

`RomLut.arch` (trimmed from `examples/rom_lut.arch`):

```arch
ram RomLut
  kind rom;
  latency 1;
  init: [0x00, 0x31, 0x5A, 0x76, 0x7F, 0x76, 0x5A, 0x31];
  param DEPTH: const = 8;
  param T: type = UInt<8>;
  port clk: in Clock<SysDomain>;

  ports rd
    addr: in UInt<3>;
    en:   in Bool;
    data: out UInt<8>;
  end ports rd
end ram RomLut
```

```
$ arch build RomLut.arch -o RomLut.sv
Wrote RomLut.sv
Wrote RomLut.archi

$ cat RomLut.archi
ram RomLut
  kind rom;
  latency 1;
  param DEPTH: const = 8;
  param T: type = UInt<8>;
  port clk: in Clock<SysDomain>;
end ram RomLut
```

The whole `ports rd ... end ports rd` group is gone. `init:` is dropped too, which is correct for an interface, but the port group is the ram's actual signal surface.

Consuming that stub via `ARCH_LIB_PATH`:

```
$ ARCH_LIB_PATH=../lib arch check Top.arch
Error: × ram `RomLut` has no port groups
```

So a separately-compiled ram is unusable: the emitted `.archi` is not a valid input to the compiler that emitted it.

## Impact

Separate compilation is the documented workflow (`doc/ARCH_HDL_Specification.md`, "Separate compilation"), and `.archi` auto-discovery is what the e203 corpus relies on. Any design that puts a `ram` behind a library boundary hits this.

It also forces a workaround in the inst-port validation added by #809: `validate_inst_ports` has to bail out entirely for an interface ram with no port groups, because otherwise every group pin (`rd_addr`, `rd_en`, `rd_data`) would be reported as an unknown port. Fixing the `.archi` emission would let that special case go away.

## Suggested fix

[...]

PR #826, opened 2026-08-09
fix(interface): stop emitting .archi for packages

### LINKED ISSUE #819: package .archi emits bodyless `function` signatures the parser rejects
## Summary

`emit_package_interface` writes each package function as a bare signature:

```
  function max(a: UInt<8>, b: UInt<8>) -> UInt<8>;
```

The parser has no production for that — a `function` must carry a body and an `end function <name>`. So a package containing any function emits an `.archi` that cannot be read back:

```
$ arch build examples/TestPkg.arch -o out.sv
Wrote TestPkg.archi

$ arch check TestPkg.archi
Error: × unexpected token: expected expression, found ;
```

## Context

Found while fixing #815 (interface stubs that do not round-trip). The other six cases in that family — ram port groups, linklist `op` blocks, regfile port arrays, arbiter custom policy, enum separators, pipeline stub validation — are fixed in the PR closing #815. This one is left out because it needs a design decision rather than a mechanical fix.

The new corpus sweep `tests/interface_roundtrip_test.rs` asserts every emitted stub re-checks; `examples/TestPkg.arch` is listed in its `known_bad` table pointing here. Removing that entry is the acceptance test for this issue.

## The decision

Two options, and they differ in more than effort:

1. **Emit the function body.** A consumer arguably needs it: `arch build` lowers a called function to a SystemVerilog `function automatic` in the caller's module, and const-folding may need to evaluate it. This requires a statement pretty-printer for function bodies, which `src/interface.rs` does not have today (it only prints types, params and port declarations).

2. **Let the parser accept a bodyless signature when reading `.archi`.** Much smaller, and consistent with how the rest of an interface works — a stub declares a surface, not an implementation. But then a caller that resolves a package function purely through `.archi` has a signature it can typecheck against and no body to emit, so codegen would need the real source anyway. That may be fine (the same is already true for module bodies) or may be a hole, depending on how package functions are lowered.

Worth deciding before implementing; option 2 is the smaller change but option 1 may be what separate compilation actually requires.

## Repro

`examples/TestPkg.arch` in-tree.

nogate-ai:
Closes #819.

## Why removal rather than a parser fix

#819 was filed as "the package stub emits bodyless `function` signatures the parser rejects." Working a concrete example showed the parse failure is the second problem, not the first: **nothing can consume a package `.archi` on any code path.**

- `use <name>;` resolves only `<name>.arch` (`src/main.rs`), never `.archi`.
- The two `.archi`-aware discovery paths key on the *construct* filename — `<inst_target>.archi` and `<bus_name>.archi` — never the package's.

Verified rather than inferred:

```
$ ls lib/          →  TestPkg.archi        # stub only
$ ARCH_LIB_PATH=$PWD/lib arch check pkg_user.arch
Error:   × undefined name: `AluOp`

$ cp TestPkg.arch lib/                     # real source alongside
$ ARCH_LIB_PATH=$PWD/lib arch check pkg_user.arch
OK: no errors
```

So teaching the parser a bodyless-`function` production would only make a file nobody reads parse.

## Can a package ever need `.archi`?

`.archi` exists for cross-file construct resolution, and a package cannot take part in it. The parser accepts only `param` / `domain` / `enum` / `struct` / `bus` / `function` inside a package:

```
$ arch check P.arch          # package containing a module
Error:   × unexpected token: expected param, domain, enum, struct, bus, or function,
  │ found module
 6 │   module Inner
```

The one package member that *does* participate in `.archi` lookup is `bus` — and it is found by its own filename, not the package's, so a package stub was never a route for it. Bus-in-package resolves today through `use` + `.arch`; confirmed with a fixture.

## Change

- `PackageDecl` drops its `iface` registration, so `emit_interface()` returns `None` (same shape `TemplateDecl` already uses).
- `emit_package_interface` and the now-dead `indent` helper are removed.
- `emit_sv` is untouched — the SV package body is unaffected.

## Result

This clears the last exception in the round-trip sweep added with #815. `tests/interface_roundtrip_test.rs` no longer carries a `known_bad` table and now asserts without qualification that **every** emitted stub re-checks.

New `packages_emit_no_archi_but_still_emit_sv_and_resolve` pins all three properties: no `.archi` written, SV package still emitted, consumer still resolves.

## Verification

- Round-trip sweep: 2 tests green, zero exceptions.
- 788 integration tests green; `cargo fmt --check` and `cargo clippy` clean.

### DIFF
--- a/doc/ARCH_HDL_Specification.md
+++ b/doc/ARCH_HDL_Specification.md
@@ -9268,6 +9268,8 @@ end module SubModule
 
 The `.archi` file is valid ARCH syntax and can be parsed by the compiler directly. It is named by **module name** (not source filename), so `fifo_async_r2w_sync.arch` (which defines `synchronizer r2w_sync`) generates `r2w_sync.archi`.
 
+**Packages emit no `.archi`.** A `package` declares types, buses and functions — never an instantiable construct — so it never takes part in the construct-name resolution `.archi` exists for, and nothing would read one: `use` resolves only `<name>.arch`, and the `.archi` lookups in §30.2 are keyed on the *construct* name rather than the enclosing package's. A consumer of a package therefore needs its `.arch`, reached with `use <Package>;`. (A bus declared inside a package is still discoverable by its own name per §30.2, since that search keys on the bus.)
+
 **30.2 Dependency Discovery**
 
 When the compiler encounters `inst sub: SubModule` and `SubModule` is not defined in the input files, it automatically searches for:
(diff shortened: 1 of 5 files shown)


nogate-ai:
Filed #828 for the modules-in-packages idea. Flagging the interaction here since it bears directly on this PR: #828's premise reverses this one's justification — if a package can declare a module, package `.archi` has to come back, and #819's bodyless-`function` problem returns with it.

That does not block merging. The removal is correct for the language as it stands today (a package cannot declare a module — the parser rejects it), and re-adding `iface` is a one-line change if #828 is ever taken up. Recording it so nobody has to re-derive why the emitter disappeared.

Issue #799, opened 2026-08-06
tests/e203: 45 of 49 checked-in C++ testbenches are stale and have not compiled since 2026-04


nogate-ai:
## Summary

Almost every `tests/e203/*_tb.cpp` file targets a class name that no construct generates any more. They were written when the e203 fixtures used PascalCase construct names (`module IfuIfetch`, `module ExuAlu`, …); the fixtures were later renamed to snake_case to match the reference RTL module names (`module e203_ifu_ifetch`), and the testbenches were never updated.

`arch sim` generates `V<ConstructName>.h`, so a tb that includes `VIfuIfetch.h` cannot compile: nothing produces that header.

**45 of 49 are stale** (audited by comparing each tb's `#include "V*.h"` against the construct name declared in its paired `.arch`):

| status | count |
|---|---|
| stale — include names a header nothing generates | 45 |
| current | 4 |

The 4 current ones: `e203_ifu_ift2icb_tb.cpp` (fixed in #785), `e203_exu_regfile_vltor_tb.cpp`, `e203_exu_wbck_vltor_tb.cpp`, and `e203_ifu_ifetch_tb.cpp` (fixed in #794).

Representative sample:

```
e203_biu_tb.cpp               VBiu.h            ->  Ve203_biu.h
e203_exu_alu_bjp_tb.cpp       VBjpUnit.h        ->  Ve203_exu_alu_bjp.h
e203_exu_alu_dpath_tb.cpp     VAluDpath.h       ->  Ve203_exu_alu_dpath.h
e203_itcm_tb.cpp              VE203Itcm.h       ->  VItcmSram.h
e203_sram_ctrl_tb.cpp         VSramCtrl.h       ->  VSramBank.h
e203_lsu_ctrl_tb.cpp          VLsuCtrl.h        ->  Ve203_lsu_ctrl.h
```

## Root cause: nothing ran them

The staleness is not the interesting part — it is that it went unnoticed for four months. Nothing in the repo executes these files:

- `tools/run_arch_regression.py` synthesizes its own smoke TB (`write_sim_smoke_tb`) and only uses a checked-in `*_tb.cpp` when the unit appears in `tests/arch_sim_manifest.json`. **No e203 unit was listed there** before #794.
- `tests/realbench/run_realbench.sh` uses RealBench's own SV testbenches, not these.
- No CI workflow references them.

So they are unreferenced files that look like coverage.

[...]

PR #843, opened 2026-08-09
fix(e203): rewire the four stale integration wrappers against the real ICB fabric

### LINKED ISSUE #798: tests/e203: four integration wrappers are wired against a superseded generation of the leaf modules
## Summary

`tests/e203/` contains two generations of fixtures that no longer fit together. The leaf modules were rewritten against the real E203 RTL (1f32bc60, 2026-05-03, "rewrite 12 modules"), but the older hand-written integration wrappers were never updated and still connect the *previous* port surfaces.

**4 of 62** e203 units fail `arch check`:

| unit | first error |
|---|---|
| `e203_core_top` | `input port commit_mret of e203_lsu_ctrl is not connected in inst lsu` |
| `e203_exu_top` | `input port i_instr of e203_exu_decode is not connected in inst dec` |
| `e203_ifu_top` | `unknown bus type ItcmIcb` (then `pc_rtvec` unconnected once the bus file is supplied) |
| `e203_soc_top` | transitively, via its `inst core: e203_core_top` |

The other 58 pass.

## This is not a one-connection fix

`arch check` stops at the first error (#750), which makes `e203_core_top` look like a single missing pin. It isn't — three of its seven insts are wired to a completely different module:

| inst | target | ports on target | connections made | connections naming a port that exists | unconnected inputs |
|---|---|---|---|---|---|
| `lsu` | `e203_lsu_ctrl` | 90 | 11 | **0** | 44 (incl. `clk`, `rst_n`) |
| `bus_` | `e203_biu` | 124 | 19 | **0** | 58 (incl. `clk`, `rst_n`) |
| `ifu` | `e203_ifu_top` | 25 | 30 | 24 | 0 |

`inst lsu` connects `addr`/`wdata`/`funct3`/`is_load`/`is_store`/`mem_*`/`load_result` — a simple flat load-store interface. The current `e203_lsu_ctrl` is the real ICB controller: `agu_icb_*`, `dtcm_icb_*`, `itcm_icb_*`, `biu_icb_*`, `nice_icb_*`, region indications. There is zero overlap, and it does not even connect the clock or reset.

`e203_exu_top` is the same story across all five of its insts (`disp` alone: 30 unconnected inputs, 88 connections that name nothing on the current `e203_exu_disp`), and `e203_ifu_top` across all three of its.

Rewiring these correctly means rebuilding the top-level integration against the real ICB fabric — a design task, not a mechanical fix.

## Why it went unnoticed

Because of #797: connections naming a nonexistent port are silently accepted. Only the *unconnected input* direction is checked, so `arch check` reports one missing pin instead of the ~120 bogus connections actually present. `e203_exu.arch` demonstrates the end state — it passes `arch check` clean while emitting SV that Verilator rejects with `PINNOTFOUND` on `o_alu`, `o_agu`, `o_alu_add`, and 21 more.

`tests/realbench/run_realbench.sh` already worked around this at line 88, building each module in isolation specifically so that "a stale wrapper module (e.g. arch-side e203_core_top.arch with broken inst port lists) no longer takes down the whole batch."

## Consequences

Beyond the 4 failing units, this keeps `tools/run_arch_regression.py` in per-file fallback mode for the whole directory: `discover_units()` groups a directory into one unit only when the whole-group `arch check` passes. That fallback used to be catastrophic (see #793); it is now merely suboptimal.

## Options

1. **Rewire against the real fabric.** Highest fidelity, real design work, and would also want `e203_soc_top` updated. Note RealBench does not ship `e203_core_top`, `e203_soc_top`, `e203_ifu_top` or `e203_exu_top` at all, so there is no reference testbench to converge against.
2. **Delete the four wrappers.** They are not part of the RealBench module list and nothing else depends on them. This makes the group `arch check` pass, restoring directory grouping.
3. **Leave them and accept 4 `skip_check_failed`.** Status quo.

Flagging rather than picking: this is a benchmark-corpus scoping decision.

## Related

- #797 — bogus inst connections pass `arch check` (the reason this drifted silently)
- #750 — only the first diagnostic is reported
- #793 — the regression-harness fallback this interacts with

nogate-ai:
## Summary

Fixes #798 by rewiring, per the maintainer's call on the issue, the four first-generation integration wrappers against the current (post-2026-05-03) leaf-module surfaces — the real ICB fabric — instead of deleting them.

- **e203_ifu_top / e203_exu_top** — rebuilt as real-fabric integrations of the current leaves, using the same wiring proven in `e203_ifu.arch` / `e203_exu.arch` (the reference RTL's own hierarchy levels; the old wrappers connected ports that no longer exist).
- **e203_core_top** — rebuilt to mirror reference `e203_core.v`: `ifu_top` + `exu_top` + `e203_lsu` + `e203_biu` wired through the real AGU/LSU/IFU ICB fabric, exporting the ITCM/DTCM/PPI/CLINT/PLIC/FIO/MEM ICBs. Width seams bridged explicitly (`agu_icb_cmd_itag` Bool→UInt<1>, `lsu_o_wbck_itag` UInt<1>→UInt<5>).
- **e203_soc_top** — rebuilt on the new core: `e203_itcm_ctrl`+`e203_itcm_ram` (TB loader enters via the ext2itcm ICB), `e203_dtcm_ctrl`+`e203_dtcm_ram`, CLINT timer behind an ICB→register adapter, and the existing (already-valid) peripheral fabric now driven by the core's dedicated ICBs: PPI→APB bridge (GPIO/UART/SPI), FIO, MEM→arbiter→SRAM, IRQ ctrl, debug module.

## Drive-by fix: latent undefined identifiers in e203_biu

Verilator lint of the emitted SV surfaced that `e203_biu.arch`'s response mux referenced five undeclared `sel_*_r` names (the declared registers are `tgt_*_r`). `arch check` accepts undefined value identifiers (#748), so this never failed before — the emitted SV was un-elaboratable. Renamed to the declared registers. This is fresh evidence for #748's severity: the corpus carried an un-lintable artifact for ~3 months.

## Notes

- The CLINT timer is the one `Reset<Sync, High>` island in an Async-Low SoC; RDC phase 2d rejects any non-identifier Reset drive, so it gets a dedicated `timer_rst` input pin.
- The 4 `*_tb.cpp` companions remain dead PascalCase-header testbenches — that's #799's separate corpus decision, untouched here.
- Regenerated the five `.sv` artifacts with the current emitter (formatting drift from the older emitter's output is expected; precedent 46db3f6d).

## Verification

- `arch check` over the whole 62-unit `tests/e203` group: clean (this restores directory grouping in `tools/run_arch_regression.py`, ending per-file fallback for the directory).
- `arch build` + full-hierarchy Verilator 5.048 `--lint-only` of `e203_soc_top` (32 modules): no errors; remaining 4 warnings pre-date this change (2 WIDTHEXPAND in `e203_exu_decode`, 2 UNOPTFLAT on the known apparent ifetch↔ift2icb loop, same topology as `e203_ifu.arch`).
- Long verification (regression harness over tests/e203) running post-open; follow-ups will be pushed if it surfaces anything.

closes #798

### DIFF
--- a/tests/e203/e203_biu.arch
+++ b/tests/e203/e203_biu.arch
@@ -261,31 +261,31 @@ module e203_biu
 
   // ── Response from selected target ────────────────────────────────────
   let rsp_valid_from_target: Bool =
-      (sel_ppi_r   & ppi_icb_rsp_valid)   |
-      (sel_clint_r & clint_icb_rsp_valid) |
-      (sel_plic_r  & plic_icb_rsp_valid)  |
-      (sel_fio_r   & fio_icb_rsp_valid)   |
-      (sel_mem_r   & mem_icb_rsp_valid);
+      (tgt_ppi_r   & ppi_icb_rsp_valid)   |
+      (tgt_clint_r & clint_icb_rsp_valid) |
+      (tgt_plic_r  & plic_icb_rsp_valid)  |
+      (tgt_fio_r   & fio_icb_rsp_valid)   |
+      (tgt_mem_r   & mem_icb_rsp_valid);
 
   let rsp_err_from_target: Bool =
-      (sel_ppi_r   & ppi_icb_rsp_err)   |
-      (sel_clint_r & clint_icb_rsp_err) |
-      (sel_plic_r  & plic_icb_rsp_err)  |
-      (sel_fio_r   & fio_icb_rsp_err)   |
-      (sel_mem_r   & mem_icb_rsp_err);
+      (tgt_ppi_r   & ppi_icb_rsp_err)   |
+      (tgt_clint_r & clint_icb_rsp_err) |
+      (tgt_plic_r  & plic_icb_rsp_err)  |
+      (tgt_fio_r   & fio_icb_rsp_err)   |
+      (tgt_mem_r   & mem_icb_rsp_err);
 
   let rsp_excl_ok_from_target: Bool =
-      (sel_ppi_r   & ppi_icb_rsp_excl_ok)   |
-      (sel_clint_r & clint_icb_rsp_excl_ok) |
-      (sel_plic_r  & plic_icb_rsp_excl_ok)  |
-      (sel_fio_r   & fio_icb_rsp_excl_ok)   |
-      (sel_mem_r   & mem_icb_rsp_excl_ok);
+      (tgt_ppi_r   & ppi_icb_rsp_excl_ok)   |
+      (tgt_clint_r & clint_icb_rsp_excl_ok) |
+      (tgt_plic_r  & plic_icb_rsp_excl_ok)  |
+      (tgt_fio_r   & fio_icb_rsp_excl_ok)   |
+      (tgt_mem_r   & mem_icb_rsp_excl_ok);
 
   let rsp_rdata_from_target: UInt<32> =
-      sel_ppi_r   ? ppi_icb_rsp_rdata   :
-      sel_clint_r ? clint_icb_rsp_rdata :
-      sel_plic_r  ? plic_icb_rsp_rdata  :
-      sel_fio_r   ? fio_icb_rsp_rdata   :
+      tgt_ppi_r   ? ppi_icb_rsp_rdata   :
+      tgt_clint_r ? clint_icb_rsp_rdata :
+      tgt_plic_r  ? plic_icb_rsp_rdata  :
+      tgt_fio_r   ? fio_icb_rsp_rdata   :
                     mem_icb_rsp_rdata;
 
   // ── Response ready from selected initiator ──────────────────────────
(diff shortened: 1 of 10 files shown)


nogate-ai:
**Long verification** (`tools/run_arch_regression.py --release --pattern 'e203/*'`):

- `tests/e203` now groups into a **single 62-file unit** again (`discover_units` whole-group check passes) — the per-file fallback the issue called out is gone.
- Grouped unit stages: `arch check` ✅ · `arch build` ✅ · Verilator lint ✅ (0 errors).
- `sim_compile` ❌ — **pre-existing, not from this PR**: the C++ sim model for `e203_itcm_ram`/`e203_dtcm_ram` (files untouched here) doesn't compile; reproduced identically on the origin/main copies with a freshly built binary. Filed as #847 (sim codegen emits an rvalue assignment for ranged part-select Vec writes). Previously invisible because these units sat outside the stale baseline while the directory was in per-file fallback.

No follow-up commits needed on this branch; the e203 unit will go fully green once #847 lands.

Issue #845, opened 2026-08-09
Pipeline stage expressions bypass the slice-base hoist (#807/#810 fixes don't apply inside `pipeline`)


nogate-ai:
## Summary

The `pipeline` construct has its own expression emitters (`src/codegen/pipeline.rs:1808`, `:1826`, `:2143`, `:2154`) that never call the `BitSlice`/`PartSelect` base hoist. So **both** arch#807's `Concat`/`Repeat` fix (PR #811) and arch#810's `FunctionCall`/`MethodCall` fix (PR #844) are inert inside a pipeline stage — the base is still emitted bare there.

Found while verifying PR #844; kept out of that PR to keep it narrowly scoped.

## Repro

```arch
domain SysDomain
  freq_mhz: 100
end domain SysDomain

pipeline Pipe810
  port clk: in Clock<SysDomain>;
  port rst: in Reset<Sync>;
  port w: in UInt<16>;
  port y: out UInt<4>;

  stage S0
    reg cap: UInt<4> reset rst => 0;
    seq on clk rising
      cap <= w.trunc<8>()[5:2];
    end seq
  end stage S0

  stage S1
    reg outr: UInt<4> reset rst => 0;
    seq on clk rising
      outr <= S0.cap;
    end seq
    comb
      y = outr;
    end comb
  end stage S1
end pipeline Pipe810
```

`arch build` emits:

```systemverilog
s0_cap <= 8'(w)[5:2];
```

which both frontends reject:

```
$ verilator --lint-only -Wno-fatal pipe810.sv
%Error: pipe810.sv:28:22: syntax error, unexpected '['

$ iverilog -g2012 -o /dev/null pipe810.sv
pipe810.sv:28: syntax error
pipe810.sv:28: error: Malformed statement
```

The `Concat`/`Repeat` bases from #807 behave the same way in this position (Icarus-only failure there; the `MethodCall` size-cast shape above fails on Verilator too).

## Why it isn't a one-line reuse of `hoist_slice_base`

The pipeline emitters rewrite identifiers with stage prefixes (`emit_pipeline_stage_expr_str` → `s0_cap`, etc.). Calling `hoist_slice_base` with the raw AST base would emit a temp assigned from **un-prefixed** identifiers, and `infer_sv_width_str`'s `$bits(...)` fallback would have the same problem. A fix needs a variant that takes the already-rewritten RHS string plus a width that is either const-folded or built from rewritten text.

## Environment

[...]

Issue #813, opened 2026-08-08
Enhancement: auto-hoist every non-portable bit-slice/part-select base — retire the `is_portable_bit_slice_base` allowlist


nogate-ai:
## Summary

`arch check` rejects `a[7:4][1:0]` and `(a - b)[3:0]` — ordinary, unambiguous ARCH — and tells the user to hand-bind the base to a `let`. Meanwhile the *single-bit* form of the identical expression, `(a - b)[3]`, is accepted and the compiler hoists the base to a temp automatically, with the spec explicitly promising "You do not need to work around this by hand."

Same base expression, same downstream SV limitation, same hoist machinery already in tree — opposite policies. This proposes retiring the `is_portable_bit_slice_base` allowlist entirely and making the automatic hoist universal, which also closes the recurring bug class that produced #650 → #807 → #808 → #810 in a single week.

## Evidence 1: the compiler already contradicts itself

Verified on a freshly built `origin/main` binary (`4c75aa30`, v0.71.0):

```arch
comb
  y = (a - b)[3];      // ✅ OK: no errors
end comb
```

emits, with no user action:

```systemverilog
  logic [($bits(a - b))-1:0] arch_idx_base_0;
  assign arch_idx_base_0 = a - b;
  assign y = arch_idx_base_0[3];
```

Change the `[3]` to `[3:0]` and the same source is refused:

```
Error:   × cannot bit-slice this expression directly; SystemVerilog backends cannot
  │ portably emit `(expr)[hi:lo]`. ... assign the expression to a typed `let`/wire
  │ first and slice the named value.
```

Chained slicing `a[7:4][1:0]` fails identically.

Spec §3.2.1 justifies the rejection with:

> The parenthesized form is not a workaround — `(a + b)[s +: 4]` and `(a[7:4])[1:0]` are themselves illegal SystemVerilog ... so silently emitting parens would just move the syntax error from `arch check` to `arch build`'s output. The compiler instead rejects the construct up front.

The premise is true; the conclusion doesn't follow. Parens-vs-reject is a false dichotomy — there is a third option, and §3.2.1 documents it *four paragraphs later* for the `Index` form. The bit-slice form just never adopted it.

## Evidence 2: the hoist is portable — dual-simulator, actually run

The #812 lesson is that portability claims need cited tools. Measured here on **Icarus Verilog 12.0 (stable, v12_0)** (`iverilog -g2012`, compile) and **Verilator 5.048** (`--lint-only`):

[...]

Issue #650, opened 2026-07-05
Make generated SystemVerilog portable to Icarus for indexed casts and sliced arithmetic


nogate-ai:
## Summary

The CVDP ARCH eval exposed generated SystemVerilog patterns that Icarus rejects even when the source ARCH type-checks. We added prompt/MCP guidance to avoid these in future agent-generated ARCH, but this should be fixed in ARCH SV emission so the compiler never emits Icarus-hostile constructs for otherwise valid ARCH.

## Problem patterns

Observed failures include generated SV that indexes the result of a cast/conversion expression, for example conceptually:

```systemverilog
$unsigned(mapped_i[...])[b]
```

and complex signed-extension/cast expressions that expand into `$bits($signed(...))` style code Icarus does not accept reliably.

A related portability issue is direct slicing/bit-selecting arithmetic expressions, conceptually:

```systemverilog
(a + b)[7:0]
(signed(a) + signed(b))[W-1:0]
```

These are legal-ish in some SystemVerilog toolchains, but have caused Icarus portability failures in benchmark evaluation.

## Desired compiler behavior

When lowering to SystemVerilog for Icarus-compatible output, avoid emitting:

- indexed casts/conversions, e.g. `$unsigned(expr)[i]`
- bit-selects/slices directly on conversion/method-call results
- bit-selects/slices directly on arithmetic expressions
- deeply nested signed-extension conversion expressions that produce complex `$bits($signed(...))` code

Instead, emit named temporaries/wires with explicit widths/types, then slice/select the named temporary. For packed output construction, prefer whole-slice or whole-lane assignments where possible.

## Current mitigation

Prompt/MCP guidance was updated to tell agents to avoid these ARCH source patterns:

- `unsigned(x)[i]`, `signed(x)[i]`, `x.sext<N>()[i]`, etc.
- `(a + b)[7:0]`, `(x - y)[i]`, `(p * q)[15:8]`
- nested `unsigned(signed(raw).sext<N>())`
- per-bit loops reading converted signed values when whole-lane assignment would work

That should improve reruns, but it is not a compiler guarantee. The durable fix belongs in SV codegen/lowering.

## Repro context

This came from the CVDP spec-to-RTL eval, especially ARCH failures in problems such as `cvdp_copilot_16qam_mapper_0001` and `cvdp_copilot_static_branch_predict_0001`, where generated SV was rejected by Icarus despite the ARCH path being otherwise plausible.

PR #853, opened 2026-08-10
fix(codegen): place slice/index hoist temps in a legal SV scope (closes #846)

### LINKED ISSUE #846: Slice/index hoist temps emitted inside `always_ff` become procedural continuous assignments
## Summary

When a hoisted slice/index base temp is generated from an expression inside a `seq` block, the `logic` declaration and its `assign` land **inside** the `always_ff` block rather than at module scope. `assign` inside a procedural block is a *procedural continuous assignment*, which is not what the codegen intends and which iverilog explicitly flags as unsupported.

Pre-existing since PR #811 (arch#807); noticed while verifying PR #844 (arch#810), which inherits the same placement. Filed separately since it is orthogonal to which base kinds get hoisted.

## Repro

```arch
module SeqCat810
  port rst: in Reset<Sync>;
  port w: in UInt<16>;
  port y: out UInt<4>;
  reg r: UInt<4> reset rst => 0;
  seq on clk rising
    r <= {w, w}[5:2];
  end seq
  comb
    y = r;
  end comb
end module SeqCat810
```

emits:

```systemverilog
always_ff @(posedge clk) begin
  if (rst) begin
    r <= 0;
  end else begin
    logic [32-1:0] arch_idx_base_0;
    assign arch_idx_base_0 = {w, w};      // <-- procedural continuous assign
    r <= arch_idx_base_0[5:2];
  end
end
```

```
$ iverilog -g2012 -o /dev/null seqcat810.sv
seqcat810.sv:17: warning: A procedural assign statement cannot be synthesized in an always_ff process.
seqcat810.sv:17: vvp.tgt sorry: procedural continuous assignments are not yet fully supported.
    The RHS of this assignment will only be evaluated once, at the time the assignment
    statement is executed.
```

Verilator compiles it silently.

## Severity: no observed wrong values, but fragile

Simulated both ways with a 2-vector TB (`w=0xBEEF` then `0x1234`, sampling after the clock edge); **both** iverilog and Verilator produce the correct results. The statement re-executes on each clock edge, so the once-per-execution evaluation iverilog describes happens to coincide with the intended sampling point.

So this is not a live miscompile — but it depends on that coincidence, it produces a synthesis-unfriendly construct (iverilog says outright it "cannot be synthesized in an always_ff process"), and it emits warnings on every build that hits it. The intended emission is a module-scope `logic` + continuous `assign`, matching what the `let`/`comb` path already produces.

## Suggested fix

`index_hoist_lines` is drained by `line()` at whatever scope is currently being emitted. It needs a module-scope drain point (or a separate pending-list that always flushes before the `always_ff`) when the hoist is triggered from inside a procedural block. Note the existing runtime-`for`-loop-variable bail already exists for a related reason and would interact with this.

## Environment

- Clean build of `origin/main` @ `e06bab7a` (arch v0.71.0); identical on PR #844's branch.
- Icarus Verilog 12.0 (stable) (v12_0); Verilator 5.048 2026-04-26.

Related: #650, #807, #810, PRs #811, #844.

nogate-ai:
## Summary

Fixes #846: the `index_hoist_lines` pending list was drained by `line()` at whatever scope happened to be emitting, so a hoist temp synthesized from an expression inside a `seq` / `comb` / `function` body had its `logic` declaration and `assign` land **inside** the enclosing `always_ff` / `always_comb` / `function automatic`. This is the *placement* half of the #650/#807/#810 hoist machinery — orthogonal to which base kinds get hoisted.

**The issue understates the severity: with two or more hoists in one procedural block, the emitted SV does not compile on either frontend.** A `logic` declaration that follows a statement inside a `begin`/`end` block is a hard syntax error, not a warning. And a hoist inside a `function automatic` body is rejected outright by both frontends — a live regression from #811, since the pre-#811 bare emission at least compiled on Verilator.

| Scope | Pre-fix emission | iverilog 12.0 | Verilator 5.048 |
|---|---|---|---|
| `seq`, one hoist | `assign` inside `always_ff` | warning + `sorry:` | silent (accidentally correct) |
| `seq`, two+ hoists | decl after a statement | **`error: Malformed statement`** | **`%Error: syntax error, unexpected IDENTIFIER`** |
| `comb` with `if` | `assign` inside `always_comb` | warning | silent |
| `function` body | `assign` to an automatic var | **`error: automatically allocated variables may not be assigned values using procedural continuous assignments`** | **`%Error: Automatic lifetime variable not allowed in continuous assignment (IEEE 1800-2023 6.21)`** |

## Change — `src/codegen/mod.rs`

`line()` gains a scope model (`HoistScope`), and the pending list becomes `Vec<HoistTemp>` (width / name / rhs) instead of `Vec<String>` of pre-rendered lines, because the legal rendering differs by scope:

- **Module / `generate` scope** — emit inline immediately before the reading statement, exactly as before. Unchanged for every existing fixture.
- **`always_ff` / `always_comb` / `always_latch` / `initial` / `final`** — splice `logic ...;` + `assign ...;` out to module scope, immediately above the block's opening line. This is what #846 asks for and what the `let`/`comb` path already produced.
- **`function automatic` body** — only the declaration moves, to the top of the body (the sole position SV permits a body declaration); the assignment stays in place as a **blocking** assignment. Module scope would be *wrong* here: the base expression may reference the function's own arguments.

Scope is tracked by `line()` itself — `begin`/`end` whole-word counting with string literals and `//` comments stripped first, plus a first-token check for the procedural openers — with an explicit push/pop around `emit_function`. Splice offsets are advanced after each insertion so a second hoist from the same block lands after the first, preserving generation order (which is dependency order: a nested hoist is created while the outer hoist's RHS is being emitted, so it is emitted ahead of it).

```systemverilog
- always_ff @(posedge clk) begin
-   if (rst) begin r <= 0; end else begin
-     logic [32-1:0] arch_idx_base_0;
-     assign arch_idx_base_0 = {w, w};      // procedural continuous assign
-     r <= arch_idx_base_0[5:2];
-   end
- end
+ logic [32-1:0] arch_idx_base_0;
+ assign arch_idx_base_0 = {w, w};
+ always_ff @(posedge clk) begin
+   if (rst) begin r <= 0; end else begin
+     r <= arch_idx_base_0[5:2];
+   end
+ end
```

### One extra fix the function-scope path required

`infer_sv_width_str`'s `Ident` arm now consults `fn_local_types` before falling back to `$bits(<name>)`. Inside a function body, arguments are not in `module_scopes`, so the temp was declared `logic [($bits(x) + $bits(q))-1:0]` — and **Icarus 12.0 computes a bogus width for `$bits(<function argument>)` in a declaration**, so the temp read back all-X. Reproduced standalone, independent of ARCH codegen:

```systemverilog
function automatic logic [3:0] PickA(input logic [7:0] x, input logic [7:0] q);
  logic [($bits(x) + $bits(q))-1:0] t;   // -> reads back X on iverilog 12.0
  t = {x, q}; return t[9:6];
endfunction
// identical body with `logic [15:0] t;` -> correct
```

`fn_local_types` is populated only while a function body is being emitted, so this lookup cannot affect module-scope width inference.

## Decision: the runtime-`for`-loop-variable bail is unchanged

`hoist_slice_base` and the `Index` hoist still skip hoisting when the base references a live `runtime_for_loop_vars` entry. I considered relaxing it and deliberately did not:

- **It is still required.** Relocating the continuous `assign` out of the `always_*` block is *exactly* what puts a loop-local `int` iterator out of scope. The bail's original reasoning survives this change unchanged for every non-function scope.
- Relaxing it would need the *other* shape — module-scope declaration plus a **blocking** assignment left inside the loop body. That works, but it trades the procedural-continuous-assign lint class for the blocking-in-`always_ff` one (`BLKSEQ`), and #846 explicitly states the intended emission is "a module-scope `logic` + continuous `assign`, matching what the `let`/`comb` path already produces". Out of scope here.
- In a `function` body the bail is moot anyway: the assignment stays in place, so a loop-local reference would remain in scope — and `emit_function_for` does not register its iterator in `runtime_for_loop_vars` to begin with. Documented on the field.

`test_index_arithmetic_base_referencing_loop_var_skips_hoist` still passes untouched.

## Determinism

Temp names are unaffected. `arch_idx_base_<n>` still comes from the same file-wide monotonic `index_hoist_counter`, allocated at the same points in the same order — the change moves *where already-numbered lines are written*, never *when a number is taken*. Confirmed by the unchanged `arch_idx_base_0` expectations in the pre-existing #650/#807/#810 shape tests.

## Verification

**Fast gate (all before this PR opened):** `cargo build --release`; full `cargo test --release --no-fail-fast` green (803 integration + 56 fp + all others, zero failures — the 2 `formal_test` Lean-replay tests pass once the gitignored `proofs/lean_thread_lowering/.lake` artifact is present); `cargo fmt --check` clean; `cargo clippy --release --all-targets` reports no errors and no new warnings in the changed regions.

**Dual-simulator, before → after.** Pre-fix binary built from `origin/main` @ `81dfdc02`, post-fix from this branch; both built fresh in this worktree.

Single hoist in a `seq` block — the symptom the issue reports:

```
BEFORE  iverilog -g2012:
  warning: A procedural assign statement cannot be synthesized in an always_ff process.
  vvp.tgt sorry: procedural continuous assignments are not yet fully supported.
AFTER   iverilog -g2012: (no diagnostic, exit 0)
```

Three hoists in one `seq` block:

```
BEFORE  iverilog:   SeqCat810.sv:22: syntax error / error: Malformed statement  (exit 4)
BEFORE  verilator:  %Error: SeqCat810.sv:22:21: syntax error, unexpected IDENTIFIER, expecting "'{"  (exit 1)
AFTER   both:       compile clean
```

Hoist inside a `function` body:

```
BEFORE  iverilog:   error: automatically allocated variables may not be assigned
                    values using procedural continuous assignments.  (exit 1)
BEFORE  verilator:  %Error: Automatic lifetime variable not allowed in continuous
                    assignment (IEEE 1800-2023 6.21): 'arch_idx_base_0'  (exit 1)
AFTER   both:       compile clean
```

**Behavioral (not just compilation).** New fixture `tests/icarus_portability/SeqSliceHoist.arch` — three hoists in one `always_ff` (`{w,w}[5:2]`, `w.trunc<8>()[7:4]`, `(w - 1)[3]`, covering both hoist call sites), one inside an `always_comb` `if` arm, one inside a `function` body. Icarus (`iverilog -g2012` + `vvp`) and Verilator (`--cc --exe --build` + run) both print `PASS` and agree on registered *and* combinational values across 2 clocked test vectors, confirming that a module-scope continuous `assign` still presents the right value at the clock edge for a `seq` context.

**Tests added** (mirroring #844's structure):

- `test_seq_hoist_temps_land_at_module_scope` — simulator-free shape check across all three hoist paths: decl + assign must precede `always_ff`, and no `logic`/`assign` line may appear inside it. This is what guards the fix on the PR gate, where neither simulator is installed.
- `test_comb_block_hoist_temps_land_at_module_scope` — same for `always_comb`, with the hoist one `begin`/`end` deeper than the opener.
- `test_function_body_hoist_declares_locally_and_assigns_blocking` — declaration at the top of the body with a concrete (non-`$bits(<arg>)`) width, blocking assignment in place, and no `assign arch_idx_base_*` anywhere.
- `test_seq_slice_hoist_behavioral_equivalence_verilator_and_iverilog` — the dual-simulator run above. It asserts the *absence* of the iverilog procedural-assign diagnostic, since the old single-hoist output already "passed" compilation. Skips gracefully when both simulators are absent.

Environment: Icarus Verilog 12.0 (stable) (v12_0); Verilator 5.048 2026-04-26.

## Note for the concurrent #845 work (`src/codegen/pipeline.rs`)

**The hoist helper's internal shape changed.** `hoist_slice_base`'s signature and behavior are unchanged (`&Expr -> Option<String>`), and it is still the right entry point, but the plumbing behind it moved:

- `index_hoist_lines: RefCell<Vec<String>>` → `index_hoist_temps: RefCell<Vec<HoistTemp>>`.
- Direct pushes of pre-rendered `format!("logic ...")` / `format!("assign ...")` strings are gone; use the new `push_hoist_temp(width, name, rhs)` helper instead. (All four in-tree push sites were converted.)
- Anything in `pipeline.rs` that emits its own `always_ff`/`always_comb` through `line()` now automatically gets the module-scope splice — no per-emitter work needed, as long as the block opener goes through `line()`.

Rebased onto `origin/main` @ `81dfdc02` immediately before pushing.

## Deliberately not done

- **spec §3.2.1 is untouched.** The fix changes nothing the section states — it describes *which* bases are hoisted and that they bind to "a compiler-generated named temporary", both unchanged. `is_portable_bit_slice_base` is likewise untouched (that is #813 P1, a spec-surface widening blocked on maintainer sign-off).
- The runtime-`for`-loop-variable bail, per the reasoning above.
- `doc/COMPILER_STATUS.md` gets a dated entry and the skill snapshot is re-synced. No `Cargo.toml` version bump.

closes #846


nogate-ai:
## Long verification (post-open)

`scripts/refactor_diff.sh origin/main --jobs 4` — **PASS**, zero diffs:

```
refactor_diff: base-ref=origin/main (81dfdc023f72b8e74ee3829a0b988c8f2b83b497)
refactor_diff: head=9e81d9c50ab321b6bfc0ba4f435d7a2464742441
refactor_diff: compared 1680 common file(s); 0 byte-differ; 0 base-only; 0 head-only.
refactor_diff: PASS
```

I had expected a legitimate non-zero diff here — the change intentionally *moves* emitted lines for any fixture that hoists inside a `seq`/`comb`/`function` body. Zero means the full regression corpus (~185 baseline units + every `tests/backend_equiv` fixture) contains **no** such site today: every existing hoist in the corpus fires from module scope, where the emission is unchanged by construction. So the blast radius on existing output is nil, and the new `tests/icarus_portability/SeqSliceHoist.arch` fixture is the only thing in-tree that exercises the fixed path.

`git merge-base --is-ancestor origin/main HEAD` held throughout, so this is a clean comparison against the same base the PR targets — not a stale-`origin/main` artifact.

Issue #854, opened 2026-08-10
pipeline stage-expression emitters fall through to the module emitter, losing stage prefixes (Repeat, Onehot, Match, Inside, StructLiteral, …)


nogate-ai:
## Summary

`src/codegen/pipeline.rs`'s two stage-expression emitters (`emit_pipeline_stage_expr_str`,
`emit_pipeline_expr_str`) handle a fixed list of `ExprKind`s and end in
`_ => self.emit_expr_str(expr)`. `emit_expr_str` is the module-level emitter: it does **not**
apply the pipeline's `<stage>_<signal>` identifier rewriting. So for any `ExprKind` that is
not in that list but contains sub-expressions, a stage signal is emitted under its bare
source name — a name that does not exist in the emitted SV.

Found while fixing #852 for `FunctionCall`, which was one instance of this:

```
      s0_r2 <= Ident8(r1);     // should be Ident8(s0_r1)
```

`FunctionCall` is fixed (PR for #852). The remaining fall-through `ExprKind`s that carry
sub-expressions and can therefore reproduce it:

- `Repeat` — `{3{stage_reg}}[...]` (`Concat` *is* handled; `Repeat` is not)
- `Onehot`
- `Match` / `ExprMatch`
- `Inside`
- `StructLiteral`
- `Range`
- `Todo`
- `LatencyAt`, `PipelinedCall`, `SvaNext`, `Implies`, `ImpliesNext` (SVA/latency forms —
  reachable from a pipeline `assert`)

## Why it is silent

`arch check` accepts the source; `arch build` succeeds; the emitted SV then fails at the
frontend with an undeclared-identifier error, or — worse — resolves to a *different*
same-named signal if one exists at module scope (e.g. a port with the same name as a stage
reg), which is a miscompile rather than an error.

## Fix shape

The same shape #845 and #852 used: don't fork the module emitter, parameterize it on its
sub-expression emitter (`hoist_slice_base_in(base, &emit, …)`,
`emit_function_call_str_in(expr, name, args, &emit_arg)`) and have both pipeline emitters
pass their stage-prefix-rewriting renderer. Doing it exhaustively (a single
`emit_expr_str_in(expr, &emit_sub)` that every context shares) is the emitter-context
abstraction tracked by #850.

## Repro (Repeat)

```arch
domain SysDomain
  freq_mhz: 100
end domain SysDomain

pipeline RepPipe
  port clk: in Clock<SysDomain>;
[...]

Issue #850, opened 2026-08-10
Enhancement: one expression emitter with explicit name-scope + hoist-sink (`EmitCtx`), gated by a shape × context × frontend conformance matrix


nogate-ai:
## Summary

Expression lowering is **re-implemented per syntactic context**, in both backends. A fix landed in
the canonical emitter does not reach the shadow copies, and nothing in CI notices. This has produced
a bug every few days for a month — #650 → #807 → #808 → #810 → #845 → #846 → #847, plus a new
`function`-body case found while writing this issue (#849) — each one a one-shape/one-context
patch on one copy of the emitter.

This proposes fixing the shape of the problem instead of the next instance: make the two things that
actually vary by context (**name resolution** and **hoist sink**) explicit parameters of a *single*
emitter, and gate the result with a generated shape × context × frontend conformance matrix.

## Evidence 1 — the duplication, counted

`ExprKind` has 59 variants. Distinct variants dispatched per emitter, on `origin/main` @ `81dfdc02`:

| Emitter | Role | distinct `ExprKind` arms |
|---|---|---|
| `src/codegen/mod.rs::emit_expr_inner` | canonical SV | **29** |
| `src/codegen/pipeline.rs::emit_pipeline_stage_expr_str` + `::emit_pipeline_expr_str` | shadow SV | 15 |
| `src/sim_codegen/expr_codegen.rs` | canonical C++ | **29** |
| `src/sim_codegen/thread_sim.rs::expr_to_cpp` | shadow C++ | 16 |
| `src/sim_codegen/pipeline.rs` | shadow C++ | 16 |

The shadows are partial re-implementations, not delegations. PR #840 ("dedup the three method-call
SV emission matches into one shared emitter") and PR #839 ("pipeline `.resize<N>()` emits bare
`N'(expr)`, matching `mod.rs`") are both hand-reconciliations of this same drift, landed the same day.

## Evidence 2 — measured, not asserted

One expression shape, `w.trunc<8>()[5:2]`, placed in nine syntactic contexts. Fresh `arch` 0.71.0
built from a clean `origin/main` worktree; `arch build ... --no-auto-asserts` (so iverilog's SVA gap
is not in play); Verilator 5.048 and Icarus Verilog 12.0.

| Context | `arch check` | Verilator | Icarus | emitted |
|---|---|---|---|---|
| module `comb` | OK | OK | OK | `arch_idx_base_0[5:2]` |
[...]

PR #855, opened 2026-08-10
fix(codegen): emit top-level `function`s into every construct's SV module (closes #852)

### LINKED ISSUE #852: A top-level `function` is never emitted into a `pipeline`'s SV module (emitted SV calls an undeclared function)
## Summary

A `function` is emitted into the SV of a `module` that calls it, but **not** into the SV of a `pipeline` that calls it. `arch check` and `arch build` both succeed; the emitted `.sv` calls a function that is nowhere declared, and every SV frontend rejects it.

Found while building the dual-simulator fixture for #845 (PR #851); it is unrelated to slicing — a bare call with no slice reproduces it — so the `FunctionCall` slice base is covered by a shape test only there, not by the dual-simulator fixture.

## Repro

```arch
domain SysDomain
  freq_mhz: 100
end domain SysDomain

function Ident8(v: UInt<8>) -> UInt<8>
  return v;
end function Ident8

pipeline FnPipe
  port rst: in Reset<Sync>;
  port a: in UInt<8>;
  port y: out UInt<8>;

  stage S0
    reg r: UInt<8> reset rst => 0;
    seq on clk rising
      r <= Ident8(a);
    end seq
    comb
      y = r;
    end comb
  end stage S0
end pipeline FnPipe
```

`arch build FnPipe.arch` succeeds. The emitted `FnPipe.sv` contains **no** `function automatic ... endfunction` at all:

```
$ grep -c function FnPipe.sv
0
```

```
$ verilator --lint-only -Wno-fatal --top-module FnPipe FnPipe.sv
%Error: FnPipe.sv:24:15: Can't find definition of task/function: 'Ident8'
   24 |       s0_r <= Ident8(a);
      |               ^~~~~~
```

`iverilog -g2012` fails the same way (`No function named 'Ident8' found in this context (FnPipe)`).

The identical `function` + call inside a `module` emits correctly — cf. `tests/icarus_portability/CallSliceHoist.arch`, whose `.sv` carries

```systemverilog
  function automatic logic [7:0] Ident8(input logic [7:0] v);
    return v;
  endfunction
```

So the function-emission step is keyed to the `module` emitter and the `pipeline` emitter never runs it.

## Severity note: it can crash iverilog outright

Combined with #845's slice-base hoist, the missing declaration makes the hoisted temp's width read `logic [$bits(Ident8(a))-1:0] arch_idx_base_0;`. Elaborating `$bits(...)` of an undeclared function **segfaults** Icarus rather than producing a diagnostic:

```
$ iverilog -g2012 -o /dev/null PipeSliceHoist.sv
Segmentation fault: 11  | /usr/local/lib/ivl/ivl ...
```

(That is an iverilog robustness bug of its own, but the input that triggers it is ours.)

## Expected

Either emit the `function automatic` into the pipeline's SV module the way the `module` emitter does, or reject the call at `arch check` with a clear diagnostic. Silently emitting SV that references an undeclared function is the one outcome that should not happen.

## Environment

- Reproduced on a freshly built binary from `origin/main` @ `81dfdc02` (arch v0.71.0) and on PR #851's branch — identical on both.
- Verilator 5.048 2026-04-26; Icarus Verilog 12.0 (stable) (v12_0).

Related: #845 / PR #851 (where this was found), #850 (the per-context emitter duplication this is another instance of).

nogate-ai:
Closes #852.

## Root cause

SystemVerilog has no free functions: a `function automatic` must live inside a module. So
`arch build` collects every top-level `Item::Function` (and `package` function) into
`Codegen::pending_functions` and copies the whole set into each emitted module — but the
step that did the copying lived **inside `emit_module`**. Every other construct has its own
emitter (`emit_pipeline`, `emit_fsm`, `emit_fifo`, …) and none of them ran it, so a
`pipeline` calling a top-level function emitted

```systemverilog
      s0_r <= Ident8(a);      // Ident8 declared nowhere in the file
```

`arch check` and `arch build` both succeeded; both frontends then rejected the output
(Verilator 5.048 `%Error: Can't find definition of task/function: 'Ident8'`, Icarus 12.0
`No function named 'Ident8' found in this context`). Combined with #845's slice-base hoist
the undeclared function also reached the temp's declared width, and Icarus **segfaults**
elaborating `logic [$bits(Ident8(a))-1:0] arch_idx_base_0;` instead of diagnosing it.

## Fix

`src/codegen/mod.rs` gains one shared `emit_pending_functions()`; `emit_module` now calls it
instead of inlining it, and it is called at the top of every construct emitter's module body:
`pipeline`, `fsm`, `fifo`, `ram`, `cam`, `counter`, `arbiter`, `regfile`, `linklist`.
Emitting the whole set (rather than only the functions a given construct references) is
`emit_module`'s long-standing behavior — an unused `function automatic` is legal, unused SV.

Internal codegen only. No user-facing syntax, type rules or documented semantics change: the
same ARCH input is accepted, the emitted SV just gains the declaration it was calling.
`is_portable_bit_slice_base` is untouched.

## Which other constructs had the same gap (all of them — checked one by one)

| construct | reachable via | before |
|---|---|---|
| `pipeline` | stage `seq` / `let` / `comb` | **broken** (the reported case) |
| `fsm` | state `comb`/`seq`, transition guards | **broken** |
| `fifo` | `assert`/`cover` | **broken** |
| `ram` | `assert`/`cover` | **broken** |
| `cam` | `assert`/`cover` | **broken** |
| `counter` | `assert`/`cover` | **broken** |
| `regfile` | `assert`/`cover` | **broken** |
| `linklist` | `assert`/`cover` | **broken** |
| `arbiter` | `policy <FnName>` hook, `assert`/`cover` | hook only |
| `synchronizer`, `clkgate` | — | **not affected, deliberately not changed** |

`pipeline` and `fsm` are reachable through their statement bodies. The other six have no
statement bodies, but every construct shares the `assert`/`cover` surface, whose expression
is emitted into its module as SVA — so `assert a: Ident8(x) == x;` on a `fifo` produced the
same undeclared-function SV. Verified for each of the eight.

`arbiter` was the one partial case: it emitted only the function named by
`policy <FnName>`. It now emits the whole set like everything else; all three arbiters in
the corpus declare exactly one function, so their emitted SV is unchanged (there is a test
pinning that the hook stays declared).

`synchronizer` and `clkgate` are **not** included and need no fix: no statement body, no
`assert`/`cover`, and only const/param expression surfaces, which cannot emit a call.
(Separately noticed while checking: a function call in a const/param position folds silently
to `0` — `param STAGES: const = Three();` emits `parameter int STAGES = 0`. Pre-existing
const-eval gap, unrelated to this PR, not fixed here.)

## Second half of the fix: pipeline call arguments lost their stage prefix

Once the function *is* declared, a second bug becomes visible in `pipeline` only. The stage
expression emitters had no `FunctionCall` arm, so a call fell through to
`_ => self.emit_expr_str(expr)` — the module-level emitter, which does not apply this
emitter's `<stage>_<signal>` rewriting:

```systemverilog
      s0_r2 <= Ident8(r1);    // `r1` does not exist; the signal is `s0_r1`
```

This was invisible before, because the SV failed on the missing declaration first — so
without fixing it, #852 would just trade "undeclared function" for "undeclared signal".
Fixed the way #845 did it rather than by forking: the `FunctionCall` arm of `emit_expr_str`
is extracted into `emit_function_call_str_in(expr, name, args, &emit_arg)`, parameterized on
its argument renderer, and both pipeline emitters pass their stage-prefix-rewriting
renderer. The `shared function` output-wire rewrite, the SVA (`past`/`rose`/`fell`) and
float (`fma`/`is_nan`) intrinsics and overload name mangling stay in exactly one place.

**Filed, not fixed here: #854.** The same `_ =>` fall-through still drops stage prefixes for
the other sub-expression-bearing `ExprKind`s — `Repeat` (`{3{r}}` → `{3{r}}`, should be
`{3{s0_r}}`), `Onehot`, `Match`/`ExprMatch`, `Inside`, `StructLiteral`, `Range`, and the
SVA/latency forms. Reproduced and written up in #854; the exhaustive fix is the
emitter-context abstraction tracked by #850, which is bigger than this bug.

## Tests

- **Simulator-free emitted-shape tests** (`tests/integration_test.rs`) — this is what guards
  the fix on the PR gate, where no simulator is installed:
  - `test_top_level_function_emitted_into_every_construct` — all eight constructs above,
    asserting `function automatic logic [7:0] Ident8(...)` + `endfunction` in each one's SV.
  - `test_arbiter_custom_policy_hook_function_still_emitted` — the hook function must not be
    lost by switching arbiter to the shared emission.
  - `test_pipeline_function_call_args_get_stage_prefix` — `Ident8(s0_r1)` present,
    `Ident8(r1)` absent.
- **Dual-simulator behavioral test + fixture** —
  `tests/icarus_portability/PipeFnCall.arch` with `tb_pipe_fn_call.sv` (Icarus) and
  `tb_pipe_fn_call_verilator.cpp` (Verilator), driven by
  `test_pipe_fn_call_behavioral_equivalence_verilator_and_iverilog` (skips gracefully when
  neither simulator is present). Six outputs: a direct call and a `FunctionCall` slice base
  in each of the three positions a pipeline emits into (stage `seq` → `always_ff`, stage
  `let` → module-scope `assign`, branching stage `comb` → `always_comb`), plus a
  never-called `Unused16` to pin that emitting the whole set is accepted by both frontends.
  Vectors and expectations are hand-derived in the TB headers.
- `PipeSliceHoist.arch`'s header comment (added by #851) said the `FunctionCall` slice base
  could not be simulated because of this bug — updated to point at the new fixture.

## Verification

Fast gate, all before this PR opened: `cargo build --release`, `cargo test --release
--no-fail-fast`, `cargo fmt --check`, `cargo clippy --release --all-targets` (no errors).

Dual-simulator, on a binary freshly built from this branch:

- `PipeFnCall`: `iverilog -g2012` + `vvp` → `PASS`; `verilator --cc --exe --build` + run →
  `PASS`. Both agree on all 6 outputs across 2 vectors.
- The `$bits(...)` segfault is gone: `iverilog -g2012 -o /dev/null` on the pipeline
  slice-base output now exits 0. Re-confirmed the crash is real by feeding Icarus the
  pre-fix shape (same SV with the three `function`/`endfunction` lines removed):
  `Segmentation fault: 11 | /usr/local/lib/ivl/ivl ...`, exit 139.
- All eight constructs' emitted SV compiles clean on both frontends (`iverilog -g2012
  -gsupported-assertions`, exit 0 each; `verilator --lint-only`, zero errors each). The
  `fsm` case additionally runs behaviorally on both (`PASS` / `PASS`).

## Notes for review

- The reported repro is confirmed on this worktree's freshly built binary before any code
  change: `grep -c function FnPipe.sv` → `0`, Verilator → `Can't find definition of
  task/function: 'Ident8'`.
- Rebased onto `origin/main` **after** PR #851 (issue #845) merged, so there is no overlap
  to reconcile in `src/codegen/pipeline.rs` / `mod.rs`.
- `doc/COMPILER_STATUS.md` gets a dated entry (+ refreshed skill snapshot). No
  `Cargo.toml` version bump. No spec change — nothing this fix touches is stated in
  `doc/ARCH_HDL_Specification.md`.


nogate-ai:
**Long verification (post-open): `scripts/refactor_diff.sh origin/main` → PASS.**

```
refactor_diff: compared 1665 common file(s); 0 byte-differ; 0 base-only; 0 head-only.
refactor_diff: PASS — zero diffs between origin/main and HEAD (6b19ca0b)
```

Zero diff is the *expected* result here even though the change adds function declarations to
emitted SV: no fixture in the regression corpus calls a top-level `function` from a
non-`module` construct (confirmed independently by grep before the run — the only corpus
files that pair a top-level `function` with a non-module construct are the three
custom-policy arbiters, `examples/arbiter_custom_hook.arch`,
`examples/nic400/Nic400ArbiterPolicy.arch` and `tests/cvdp/bus_arbiter_rr.arch`, and each
declares exactly one function — the hook — so switching arbiter from "emit the hook" to
"emit the whole set" is byte-identical for all three).

PR #857, opened 2026-08-10
fix(sim): lower runtime-bound ranged part-selects correctly (closes #847)

### LINKED ISSUE #847: sim_codegen: ranged part-select assignment to Vec element emits non-compilable C++ (rvalue assignment, wrong width)
## Repro

`tests/e203/e203_itcm_ram.arch` (unmodified, on current main) — byte-masked RAM write:

```arch
reg mem: Vec<UInt<64>, 8192> reset none;
...
if cs & we
  for i in 0..7
    if wem[i:i]
      mem[addr][(i * 8 + 7):(i * 8)] <= din[(i * 8 + 7):(i * 8)];
    end if
  end for
end if
```

```bash
arch sim tests/e203/e203_itcm_ram.arch --outdir /tmp/out
c++ -std=c++17 -fsyntax-only /tmp/out/Ve203_itcm_ram.cpp
```

emits (iteration i=4 shown):

```cpp
if ((wem >> 32) & 0x1ULL) {
  (((_ARCH_BCHK((addr), 8192, "mem"), _mem[addr]) >> 32) & 0x1ULL)  = ((din >> 32) & 0x1ULL);
```

Three defects in one line:
1. **Assignment target is an rvalue** — the LHS part-select is emitted as a shift-and-mask *read* expression, not a read-modify-write; clang/gcc reject with `expression is not assignable`.
2. **Width is wrong** — the source selects an 8-bit byte lane `[(i*8+7):(i*8)]`, the emitted mask is `0x1ULL` (1 bit).
3. **The `wem[i:i]` condition inherits the wrong shift** — `wem >> 32` on a UInt<8> (shift-count-overflow warning); the loop iterates i=0..7 but the emitted shifts are the *byte-lane* offsets (0,8,16,24,32,...), so the mask bit index and the data offset got conflated.

`arch check` and `arch build` both accept the construct (the emitted SV is fine); only the C++ sim model is broken. `e203_dtcm_ram.arch` (UInt<32>/wem 4) fails identically.

## Impact

Found by the tests/e203 regression run for PR #843: with the four stale wrappers rewired, `tests/e203` groups into a single 62-file unit again, and that unit now fails at `sim_compile` on these two files. Pre-existing — reproduces on the origin/main copies with a freshly built binary; previously invisible because per-file fallback (#798) plus the stale regression baseline (#793 thread) kept these units out of the gate.

Likely locus: sim codegen for seq assignment where the target is `Vec[elem][hi:lo]` with loop-iterator-dependent bounds — the part-select lowering emits the read form of the slice on the LHS instead of the RMW pattern, with the 1-bit special case's mask.

nogate-ai:
Closes #847.

## Problem

The C++ sim codegen emitted non-compilable code for seq assignments whose target is a ranged part-select with runtime (loop-variable) bounds:

```arch
for i in 0..7
  if wem[i:i]
    mem[addr][(i * 8 + 7):(i * 8)] <= din[(i * 8 + 7):(i * 8)];
  end if
end for
```

produced

```cpp
if ((wem >> 32) & 0x1ULL) {
  (((..., _mem[addr]) >> 32) & 0x1ULL)  = ((din >> 32) & 0x1ULL);   // rvalue LHS — does not compile
}
```

Three symptoms, one root cause: every slice bound went through `eval_width_in`, whose 32-default silently const-folds anything it can't evaluate — so the LHS became a read expression, the lane mask collapsed to 1 bit, and the `wem[i:i]` shift became a literal 32. The emitted SV from `arch build` was already correct; only the sim model was broken. `tests/e203/e203_itcm_ram.arch` and `e203_dtcm_ram.arch` both failed sim-compile this way (found during #843's long verify).

## Fix (internal codegen only — no user-facing syntax/semantics change)

- **`expr_codegen.rs` read path**: `BitSlice` with non-const bounds now emits a symbolic shift by the runtime lo bound, with the mask taken from a structural slice-width analysis (`[x:x]` → 1, `[x+k:x]` → k+1, `[x:x-k]` → k+1, plus the existing param-aware const fold). When the width itself is runtime (e.g. `d[i:0]`), a new `_arch_slice_mask` runtime helper computes it. Runtime bounds also get an `_ARCH_BCHK` MSB check, mirroring the variable part-select policy (typecheck can only verify compile-time bounds). **Constant-bound slices emit exactly as before.**
- **`infer_expr_width`**: uses the same structural width, so `din[i*8+7:i*8]` reports 8 bits instead of 1.
- **`stmt_codegen.rs` LHS arm**: the bit-slice read-modify-write lowering now (a) fires for Vec-element bases — `infer_expr_width` returned 0 for `mem[addr]` because `build_widths` registers Vec regs with the scalar default 32, so the arm was skipped and the broken generic path ran — and (b) emits the shift symbolically for runtime bounds. The RMW targets the next-cycle shadow (`_n_mem[addr]`), which is memcpy'd from the current state at cycle entry, so partial-lane nonblocking semantics are exact.
- **`pybind.rs`** (`verilated.h` template): adds `_arch_slice_mask`.

Emitted code for the repro now:

```cpp
if (_ARCH_BCHK((i), 8, "wem[hi:lo]"), (((wem) >> (i)) & 0x1ULL)) {
  _ARCH_BCHK((...i * 8 + 7...), 64, "mem[hi:lo]");
  (..., _n_mem[addr]) = ((..., _n_mem[addr]) & ~(uint64_t(0xFFULL) << (i * 8...))) | (...);
}
```

## Verification

Fast gate (pre-PR):
- `cargo build --release` + full `cargo test --release`: green. (Two pre-existing failures in `formal_test` — `construct_proof_lean_*` — are local Lean-toolchain environment issues; confirmed identical on unmodified `origin/main`.)
- `arch sim tests/e203/e203_itcm_ram.arch` and `e203_dtcm_ram.arch` → `c++ -std=c++17 -fsyntax-only` clean (previously: "expression is not assignable").
- Functional check: byte-masked write TB against the ITCM model — masked lanes written, unmasked lanes preserved, `wem=0` is a no-op.

New regression coverage:
- `tests/sim_ranged_slice_regression.arch` + `_tb.cpp` — distilled byte-masked RAM, registered in `tests/arch_sim_manifest.json`.
- `tests/sim_ranged_slice_test.rs` — (1) end-to-end `arch sim --tb` semantics run; (2) e203 ITCM model host-C++ syntax check + shape guards (8-bit lane mask present, no `wem >> 32`).

Long verify (post-PR, running): `tools/run_arch_regression.py --release --pattern 'e203/*'` — will report results in a follow-up comment.

### DIFF
--- a/src/sim_codegen/expr_codegen.rs
+++ b/src/sim_codegen/expr_codegen.rs
@@ -42,6 +42,126 @@ pub(super) fn cast_to_signed_bits(expr: &str, bits: u32) -> String {
     }
 }
 
+/// Structural equality over the small expression grammar that appears in
+/// bit-slice bounds (idents, integer literals, arithmetic over them).
+/// Conservative: false on anything unrecognized.
+fn slice_bound_expr_eq(a: &Expr, b: &Expr) -> bool {
+    fn lit_u64(l: &LitKind) -> Option<u64> {
+        match l {
+            LitKind::Dec(v) | LitKind::Hex(v) | LitKind::Bin(v) => Some(*v),
+            LitKind::Sized(_, v) | LitKind::ParamSized(_, v) => Some(*v),
+            LitKind::Float(_) | LitKind::TypedFloat(_, _) => None,
+        }
+    }
+    match (&a.kind, &b.kind) {
+        (ExprKind::Ident(x), ExprKind::Ident(y)) => x == y,
+        (ExprKind::Literal(x), ExprKind::Literal(y)) => {
+            matches!((lit_u64(x), lit_u64(y)), (Some(u), Some(v)) if u == v)
+        }
+        (ExprKind::Binary(o1, l1, r1), ExprKind::Binary(o2, l2, r2)) => {
+            o1 == o2 && slice_bound_expr_eq(l1, l2) && slice_bound_expr_eq(r1, r2)
+        }
+        (ExprKind::Unary(o1, e1), ExprKind::Unary(o2, e2)) => {
+            o1 == o2 && slice_bound_expr_eq(e1, e2)
+        }
+        (ExprKind::Clog2(x), ExprKind::Clog2(y)) => slice_bound_expr_eq(x, y),
+        _ => false,
+    }
+}
+
+/// Constant width of a `[hi:lo]` slice when derivable (arch#847):
+/// - both bounds const-fold (params included) → `hi - lo + 1`
+/// - `hi` ≡ `lo` structurally → 1 (e.g. `wem[i:i]` in a `for`)
+/// - `hi` = `lo + k` / `k + lo` with `k` const → `k + 1`
+///   (e.g. `din[i*8+7 : i*8]`)
+/// - `lo` = `hi - k` with `k` const → `k + 1`
+/// Returns None when the width genuinely depends on a runtime value
+/// (e.g. `d[i:0]`) — callers fall back to a runtime mask.
+pub(super) fn try_slice_const_width(hi: &Expr, lo: &Expr, ctx: &Ctx) -> Option<u32> {
+    if let (Some(h), Some(l)) = (
+        try_eval_const_expr_with_params(hi, ctx.params),
+        try_eval_const_expr_with_params(lo, ctx.params),
+    ) {
+        return h.checked_sub(l).map(|d| d as u32 + 1);
+    }
+    if slice_bound_expr_eq(hi, lo) {
+        return Some(1);
+    }
+    if let ExprKind::Binary(BinOp::Add, a, b) = &hi.kind {
+        if slice_bound_expr_eq(a, lo) {
+            if let Some(k) = try_eval_const_expr_with_params(b, ctx.params) {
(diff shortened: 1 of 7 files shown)


nogate-ai:
Long verification complete (branch rebased onto current `main` to pick up #843's wrapper rewires):

- `tools/run_arch_regression.py --release --pattern 'e203/*'` → **pass** (the grouped `e203` unit now gets past `sim_compile` and completes end-to-end; on unmodified `main` the same sweep fails at `sim_compile` — the #847 bug).
- Full `cargo test --release --no-fail-fast` on the rebased branch: all binaries green (1,077 tests) except the two known `construct_proof_lean_*` failures in `formal_test`, which reproduce identically on unmodified `origin/main` (local Lean toolchain environment, unrelated).

Note for reviewers: pre-rebase, the per-file e203 sweep showed `e203_biu` failing verilator_lint and the three `*_top` wrappers failing `arch check` — those were the stale wrappers #843 fixed on `main`, present at the old branch base, and are gone after the rebase.

PR #860, opened 2026-08-10
fix(sim): resolve Vec element widths from declared types (closes #858, #859)

### LINKED ISSUE #858: sim: Vec element width inferred as widths[name]/count, but build_widths stores 32 for Vec types — wrong concat shifts, wrong/skipped _ARCH_BCHK bounds
## Bug

`infer_expr_width`'s `ExprKind::Index` arm (src/sim_codegen/expr_codegen.rs) computes a Vec element's width as `ctx.widths[name] / count`. But `build_widths` (src/sim_codegen/width.rs) registers Vec-typed regs/ports/wires via `type_bits_te_with_params`, which falls through to the scalar default **32** for `TypeExpr::Vec` — not the packed total. So:

- `Vec<UInt<32>, 4>[i]` infers **8** instead of 32
- `Vec<UInt<64>, 8>[i]` infers **4** instead of 64
- `Vec<UInt<64>, 8192>[i]` infers **0** (count > 32)

## Symptoms (reproduced on a fresh binary at 30446202)

Distilled fixture: reg `v: Vec<UInt<32>, 4>`, `w: Vec<UInt<64>, 8>`:

1. **Concat shift positions wrong**: `{v[1], v[0]}` emits `v[1] << 8`; result `0x000000deafbeff55` instead of `0xDEADBEEFAAAA5555`. `{w[1], w[0]}` (128-bit) is fully scrambled.
2. **Runtime bit-select bounds check wrong**: `w[0][bit_sel]` with `bit_sel = 60` hard-aborts `ARCH-ERROR: <bitsel>[i]: index 60 out of bounds [0..4)` — a legal read of bit 60 of a 64-bit element. For large counts the inferred width 0 makes the `base_w > 0` gate silently *skip* the check instead.
3. Anything downstream of element widths is poisoned: shift positions, `~` masking, wide-path classification. The bit-slice LHS path was already worked around locally in stmt_codegen.rs `slice_lhs_base_width` during the #847 fix (PR #857).

## Fix direction

Resolve the element width from `ctx.decl_types` (declared `TypeExpr::Vec` element type, param-aware) in the Index arm; keep total/count as the fallback for names whose widths-map entry IS a packed total (inst-output Vec wires, mod.rs `elem_bits * count`) and for ctxs built without decl_types (pipeline stage bodies). The #847 local workaround then folds into the shared path.

The emitted SV from `arch build` is unaffected (SV codegen has its own width resolution) — this is sim-model-only.

### LINKED ISSUE #859: sim: let-bound 65–128-bit output port truncates through uint64_t (missing _arch_u128_to_vl conversion)
Found while fixing #858, but independent of Vecs.

`let y = {a, b};` with `y: out UInt<128>` (any 65–128-bit output port bound via a module-scope `let`) emits in the sim model:

```cpp
y = (((_arch_u128)(uint64_t)(b) << 0) | ((_arch_u128)(uint64_t)(a) << 64));
```

`y` is `VlWide<4>`, the RHS is `_arch_u128`; the assignment resolves through a `uint64_t` conversion, so words 2–3 (bits 64–127) are silently zeroed. Reproduced on a fresh binary at 30446202 with a plain scalar concat — no Vec involved.

The comb-assign path in `stmt_codegen.rs` already handles this case correctly via `_arch_u128_to_vl(rhs, port._data, words)` (with a `> 128` direct-assign arm); the `let`-binding emission in `sim_codegen/mod.rs` predates that conversion and never got it.

The emitted SV from `arch build` is unaffected — sim-model-only.

nogate-ai:
## Summary

Fixes two sim-model-only miscompiles (emitted SV from `arch build` is unaffected — SV codegen has its own width resolution):

**#858 — Vec element width inference.** `infer_expr_width`'s `Index` arm computed a Vec element's width as `widths[name] / count`, but `build_widths` registers Vec-typed regs/ports at the scalar default 32 (`type_bits_te_with_params` has no Vec arm). So `Vec<UInt<32>,4>[i]` inferred 8, `Vec<UInt<64>,8>[i]` inferred 4, and any count > 32 inferred 0. Reproduced on a fresh binary: `{v[1], v[0]}` emitted `v[1] << 8` (result `0x000000deafbeff55` instead of `0xDEADBEEFAAAA5555`), and a legal read of bit 60 of a 64-bit element hard-aborted with `ARCH-ERROR: index 60 out of bounds [0..4)`; counts > 32 silently *skipped* the bounds check instead (`base_w > 0` gate).

**#859 — wide-`let` output-port truncation** (found while testing, Vec-independent). `let y = {a, b};` with `y: out UInt<128>` assigned an `_arch_u128` RHS straight into a `VlWide<4>` port, truncating through `uint64_t` — bits 64–127 silently zeroed.

## Fix

- `expr_codegen`: the `Index` arm resolves the element width from the declared `Vec<T,N>` type (`sim_expr_decl_type`, param-aware) via a new `width.rs` helper `scalar_type_bits_with_params` that returns `None` for aggregate types instead of inventing a 32 default. The `total/count` computation stays as **fallback** for the names whose widths-map entry IS a packed total — inst-output Vec wires (`elem_bits * count` in mod.rs) — and for ctxs built without `decl_types` (pipeline stage bodies), so those paths are byte-identical.
- `stmt_codegen`: the #847 local `slice_lhs_base_width` workaround folds into the now-correct shared path (both #847 regression tests still pass).
- `mod.rs` let emission: apply the same `_arch_u128_to_vl` conversion as `stmt_codegen`'s comb arm for 65–128-bit let-bound output ports (>128-bit stays direct VlWide assign).

## Call-site audit

Consumers of `infer_expr_width` on Index expressions all improve or are unchanged: concat shift positions, Shl/Shr widths, `~` masking, wide-path classification, `_ARCH_BCHK` gating/bounds, and the bit-slice LHS RMW arm. Known remaining niche: a Vec-typed **bus field** whose element width is an unresolvable bus param still degrades (to 32, vs `32/count` pre-fix — no worse); noted rather than fixed since the bus decl_types entries are pre-substitution.

## Verification (fast gate — green before PR)

- `cargo build --release` clean; full `cargo test --release`: **19/19 targets, 0 failures** (incl. both #847 tests with the workaround removed).
- New regression fixture `sim_vec_elem_width_regression` passes 6/6 end-to-end (32/64-bit element concats into 64/128-bit results, runtime bit-select at 60/63, wide-let conversion) plus codegen-shape assertions (`_ARCH_BCHK(..., 64, ...)`, `<< 64`, `_arch_u128_to_vl`). Pre-fix binary fails 4 of 6 and aborts on the bit-select.

Long verification (full regression sweep) runs post-open per fast-gate-before-PR ordering; follow-ups will be pushed here if it surfaces anything.

closes #858
closes #859

### DIFF
--- a/src/sim_codegen/expr_codegen.rs
+++ b/src/sim_codegen/expr_codegen.rs
@@ -608,7 +608,24 @@ pub(super) fn infer_expr_width(expr: &Expr, ctx: &Ctx) -> u32 {
                     .vec_names
                     .map_or(false, |s| s.contains(base_name.as_str()))
                 {
-                    // Vec element width: total port/reg/field width / element count.
+                    // Preferred: element width from the declared `Vec<T, N>`
+                    // type, param-aware. The total/count fallback below is
+                    // wrong for declared Vecs — `build_widths` registers
+                    // Vec-typed regs/ports at the scalar default 32, not the
+                    // packed total, so `Vec<UInt<32>,4>[i]` computed 8 and
+                    // any count > 32 computed 0 (arch#858).
+                    if let Some(elem_ty) = sim_expr_decl_type(expr, ctx) {
+                        if let Some(w) = scalar_type_bits_with_params(&elem_ty, ctx.params) {
+                            if w > 0 {
+                                return w;
+                            }
+                        }
+                    }
+                    // Fallback: total width / element count. Correct for the
+                    // names whose widths-map entry IS a packed total —
+                    // inst-output Vec wires (registered as elem_bits * count
+                    // in mod.rs) — and for ctxs built without decl_types
+                    // (e.g. pipeline stage bodies).
                     let total = ctx.widths.get(base_name.as_str()).copied().unwrap_or(0);
                     let count = ctx
                         .vec_sizes
--- a/src/sim_codegen/mod.rs
+++ b/src/sim_codegen/mod.rs
@@ -3228,14 +3228,32 @@ impl<'a> SimCodegen<'a> {
                 if l.ty.is_none() {
                     // ty=None: assignment to existing port or wire
                     let name = &l.name.name;
-                    let target = if port_names.contains(name) {
-                        // Output port — public field, plain name
-                        name.clone()
+                    if port_names.contains(name) {
+                        // Output port — public field, plain name. Wide ports
+                        // need the same 65–128-bit conversion stmt_codegen's
+                        // comb arm applies: expression-context RHS is
+                        // _arch_u128, the port is VlWide<ceil(W/32)>, and a
+                        // bare assignment truncates through uint64_t
+                        // (`let y = {a, b};` with y: out UInt<128> dropped
+                        // the high word pair — found while fixing arch#858).
+                        if wide_names.contains(name.as_str()) {
+                            let bits = widths.get(name.as_str()).copied().unwrap_or(0);
+                            if bits > 128 {
+                                // >128 bits: both sides are VlWide<N> — direct assign.
+                                cpp.push_str(&format!("  {name} = {val};\n"));
+                            } else {
+                                cpp.push_str(&format!(
+                                    "  _arch_u128_to_vl({val}, {name}._data, {});\n",
+                                    wide_words(bits)
+                                ));
+                            }
+                        } else {
+                            cpp.push_str(&format!("  {name} = {val};\n"));
+                        }
                     } else {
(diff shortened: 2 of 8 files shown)


nogate-ai:
Long verification complete and green: `nightly_equivalence.py` main sweep **178/178 units passed on first pass** (check → build → Verilator lint → native sim with TBs, no flake retries needed), and **backend_equiv 16/16** torture fixtures pass. No divergences.

Issue #427, opened 2026-05-26
sim_codegen: `Concat` pack uses wrong shift offsets when bus-port signal width comes from module-param-substituted bus-alias params


nogate-ai:
## Summary

When a `Concat({...})` reads bus-port signals AND one of the bus's per-signal widths is determined by a **module-param-substituted bus alias param**, the C++ sim codegen emits wrong shift offsets in the pack expression. Unpack (bit-slice) is unaffected. SV codegen is unaffected (it's parametric). Hit while building NIC-400 per-edge register slices (where 5 channels concat AR/R/AW/W/B payloads into a wide `UInt<PAYLOAD_W>` for the generic `RegSliceChannel`).

This is the seventh thread/codegen-lowering issue uncovered through this user's NIC-400 work. Sibling pattern to #423 — a bus-param substitution that works in some emitters drops out in another.

## Minimal repro

`/tmp/ConcatWidthBug.arch`:

```arch
bus MiniAxi
  param ADDR_W: const = 32;
  param ID_W:   const = 1;
  ar_valid: out Bool;
  ar_ready: in  Bool;
  ar_addr:  out UInt<ADDR_W>;
  ar_id:    out UInt<ID_W>;
end bus MiniAxi

module ConcatWidthBug
  param ADDR_WIDTH: const = 32;
  param ID_W: const = 3;
  port clk: in Clock<SysDomain>;
  port rst: in Reset<Async, Low>;
  port up: target MiniAxi<ADDR_W=ADDR_WIDTH, ID_W=ID_W>;

  param PAYLOAD_W: const = ADDR_WIDTH + ID_W;
  let p: UInt<PAYLOAD_W> = {up.ar_addr, up.ar_id};
  ...
end module
```

Build:
```
cargo run --quiet --bin arch -- sim /tmp/ConcatWidthBug.arch -o /tmp/sim
grep "_let_p" /tmp/sim/*.cpp
```

## Expected

`up.ar_id` is `UInt<3>` (ID_W=3 bound via module param). `{up.ar_addr, up.ar_id}` should pack with addr in the high bits, id in the low 3:
```cpp
_let_p = (((uint64_t)(up_ar_id) << 0) | ((uint64_t)(up_ar_addr) << 3));
```

## Actual

```cpp
_let_p = (((uint64_t)(up_ar_id) << 0) | ((uint64_t)(up_ar_addr) << 32));
//                                                                  ^^ wrong: 32, not 3
```

[...]

PR #863, opened 2026-08-10
fix(sim): substitute bus params into decl_types; warn on unresolvable Vec element widths (closes #862)

### LINKED ISSUE #862: sim: Vec-typed bus field element width silently degrades — decl_types stores pre-substitution bus signal types
Follow-up to #858 (PR #860), promoting the "known accepted niche" from that PR to a fixed bug per review feedback: **the degradation must not be silent**.

## Bug

The sim codegen's `decl_types` map stores bus-port signal types straight from the bus *declaration* (`bd.signals`), without substituting bus params from the port-site binding. So for

```arch
bus VecStream
  param W: const = 8;
  data:  in Vec<UInt<W>, 4>;
end bus VecStream

module M
  port s: initiator VecStream<W=16>;
  port cat: out UInt<32>;
  let cat = {s.data[1], s.data[0]};
end module M
```

the element type resolves as `UInt<W>` where `W` is meaningless at module scope; the width evaluator's silent 32-fallback then shifts `s.data[1]` by 32 — out of the 32-bit result entirely (`cat == 0x0000beef` instead of `0xDEADBEEF`). Reproduced on a binary at PR #860's HEAD. No warning is emitted anywhere.

## Fix

1. Substitute bus params into the bus-port `decl_types` entries with the same recipe `flatten_bus_port_with_dir` already uses (defaults overridden by the binding, `effective_signals` + `subst_type_expr_sim`) — the widths then resolve correctly.
2. Make `scalar_type_bits_with_params` return `None` for UInt/SInt widths the const evaluator cannot fold (instead of `eval_width`'s silent 32), and have `infer_expr_width`'s Index arm emit a warning before using the packed-width/count fallback when a declared *scalar* element width fails to fold — so any remaining degradation is loud.

Sim-model-only; `arch build` SV output unaffected.

nogate-ai:
## Summary

Follow-up to #858 / PR #860, promoting its documented "known accepted niche" to a fix per review feedback: **a Vec-typed bus field's element width must not silently degrade.**

Reproduced at PR #860's HEAD: a `Vec<UInt<W>, 4>` bus signal bound at the port site as `VecStream<W=16>` resolved its element type as `UInt<W>` with `W` meaningless at module scope; the width evaluator's silent 32-fallback then shifted `{s.data[1], s.data[0]}`'s high element clean out of the 32-bit result (`0x0000beef` instead of `0xDEADBEEF`) — with no diagnostic anywhere.

## Fix

1. **Resolve correctly** (mod.rs): bus-port `decl_types` entries now carry bus-param-substituted signal types, using the exact recipe `flatten_bus_port_with_dir` already uses (bus defaults overridden by the port-site binding, `effective_signals` + `subst_type_expr_sim`). Substituted types may still reference enclosing-module params — those resolve later through the param-aware width helpers, matching #427's established pattern.
2. **Never guess silently** (width.rs): `scalar_type_bits_with_params` now returns `None` for UInt/SInt widths the const evaluator cannot fold, instead of `eval_width`'s silent 32. No resolution regression: the const evaluator folds a strict superset of `eval_width`'s shapes (literals, params, `$clog2`, arithmetic, ternary).
3. **Warn on fallback** (expr_codegen): when a declared *scalar* Vec element width fails to fold, the Index arm emits a warning (matching the existing unknown-ident warning style) before using the packed-width/count fallback. Aggregate elements (2D Vec, struct) fall through without noise — the fallback is the intended path for those.

Sim-model-only; `arch build` SV output unaffected.

## Verification (fast gate — green before PR)

- Full `cargo test --release`: all targets green (incl. the new `bus_vec_field_elem_width_sim`, which also asserts the warning does **not** fire when the binding resolves).
- New fixture `sim_bus_vec_elem_width` verified **failing at PR #860's HEAD** (`0x0000beef`, both checks) and passing post-fix (`0xdeadbeef`), registered in `tests/arch_sim_manifest.json` so the nightly equivalence sweep exercises it.

Long verification (full 179-unit regression sweep + backend_equiv) runs post-open per fast-gate-before-PR ordering; follow-ups will be pushed here if it surfaces anything.

closes #862

### DIFF
--- a/src/sim_codegen/expr_codegen.rs
+++ b/src/sim_codegen/expr_codegen.rs
@@ -619,6 +619,20 @@ pub(super) fn infer_expr_width(expr: &Expr, ctx: &Ctx) -> u32 {
                             if w > 0 {
                                 return w;
                             }
+                        } else if matches!(elem_ty, TypeExpr::UInt(_) | TypeExpr::SInt(_)) {
+                            // A declared scalar element whose width won't
+                            // const-fold means the total/count fallback below
+                            // is a guess, not a resolution — say so instead
+                            // of silently degrading (aggregate elements — 2D
+                            // Vec, struct — fall through without noise; the
+                            // fallback is the intended path for those).
+                            eprintln!(
+                                "warning: sim codegen: element width of Vec \
+                                 '{base_name}' could not be resolved from its \
+                                 declared type; falling back to packed-width/\
+                                 count inference — concat / shift positions \
+                                 derived from this may be incorrect"
+                            );
                         }
                     }
                     // Fallback: total width / element count. Correct for the
--- a/src/sim_codegen/mod.rs
+++ b/src/sim_codegen/mod.rs
@@ -506,12 +506,25 @@ impl<'a> SimCodegen<'a> {
             // Bus ports contribute compound "port.signal" keys with each
             // signal's declared type, so float bus fields (`s.data + ...`)
             // dispatch float ops instead of integer ops on the bit pattern.
+            // Bus params are substituted (defaults overridden by the
+            // port-site binding, same recipe as flatten_bus_port_with_dir):
+            // the raw declaration types carry bus-param idents that are
+            // meaningless at module scope, which silently degraded any
+            // width resolved through decl_types — e.g. a Vec<UInt<W>,N>
+            // signal's element width (arch#858).
             if let Some(bi) = &p.bus_info {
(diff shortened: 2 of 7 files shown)


nogate-ai:
Long verification complete and green: `nightly_equivalence.py` main sweep 178/178 units, no divergences; backend_equiv torture fixtures PASS.

Issue #829, opened 2026-08-09
Migrate the remaining ~40 float-format registration sites onto fp_format::FORMATS


nogate-ai:
Follow-up to Phase 0 of the approved block-format design (PR #824 §5). Phase 0 introduced `src/fp_format.rs` as the canonical format table and moved the two **silently wrong** wildcards onto it. The remaining hand-written maps still derive format facts independently and should be migrated.

## Already done (Phase 0)

- `formal::float_tag_width` — was `match tag { "f32" => 32, "bf16" => 16, _ => 8 }`; a 4- or 6-bit format silently took 8.
- `elaborate` literal-overflow diagnostic — was `E4m3 => ("FP8E4M3", 448.0), _ => ("FP8E5M2", 57344.0)`; every other format misreported as FP8E5M2.
- `Ty::is_float` split into carrier vs `Ty::is_float_arith`, with the `+ - *` / `fma` / `is_nan` gates routed through the latter.

## Still hand-written (this issue)

**Silently-wrong-for-a-new-format (highest value):**
- The `is_nan` bit-field tests, hand-written per tag with an `_ =>` **f32 fallback**, in `src/codegen/mod.rs`, `src/formal.rs`, and the C++ runtime in `src/sim_codegen/pybind.rs`. Note E4M3 is *not* the IEEE pattern (OCP: sole NaN `0x7F`), so a generic generator needs a per-format **NaN rule** in the descriptor, not just exp/mant bit counts.
- Three `unwrap_or("f32")` tag fallbacks (`formal.rs`, `codegen/mod.rs`). These appear unreachable today — `expr_float_tag` failing implies a non-float, which typecheck already rejects — but they would silently mis-dispatch a new format. Needs a per-site reachability argument before changing, which is why Phase 0 left them alone.

**Mechanical duplication:**
- `codegen/mod.rs`'s `param_float_fmt` contains **three copies** of the same 4-arm map (module / fsm / package).
- **Six independent `float_names` builders**, one per construct (module, fsm, pipeline, functions, …).
- `sim_codegen/const_eval.rs` uses four independent `matches!` booleans rather than an exhaustive match — invisible to the compiler when a variant is added.

[...]

PR #872, opened 2026-08-10
feat(fp): ScaledVec<Elem,N,Scale> block-scaled type + package-scope type aliases

nogate-ai:
Phase 2a of the approved MX/NVFP4 design (#824): the `ScaledVec` type, its layout, and the `split` port form.

**Not in this PR:** `scaled_quantize` / `scaled_dequantize` / `scaled_dot`, and the `UE4M3` scale. Those are phases 2b/3/5.

## The type

```arch
package MxFormats
  type MXFP4 = ScaledVec<FP4E2M1, 32, E8M0>;
  type MXFP8 = ScaledVec<FP8E4M3, 32, E8M0>;
end package MxFormats

module Consumer
  port a: in MXFP4;         // one packed logic [135:0]
  port c: in split MXFP4;   // c_scale [7:0] + c_elems [127:0]
end module Consumer
```

A block is **one packed word** `{scale, P[N-1], …, P[0]}` of width `scale_w + N*elem_w` — 136 bits for MXFP4, 264 for MXFP8. Unlike `Vec<T,N>` it is not an aggregate, so it flows through the pipeline as a wide scalar, and `arch formal` accepts it directly because it already fits the `(_ BitVec W)`-per-signal model.

## Two rejections that close *silent* paths

OCP MX §6 defines exactly one operation on blocks — the dot product — and no add, multiply, or compare. These are explicit arms rather than merely-absent features because in both cases the fallthrough was a wrong answer, not an error:

- **`a[0]`** fell through `ExprKind::Index`'s `_ => Ty::UInt(1)`, typing a 136-bit block read as one bit.
- **`a + b`** fell past the float guards into the **integer** arms, adding two packed `{scale, elements}` words — carrying element overflow straight into the scale field.

```
× operator `+` is not supported on ScaledVec<FP4E2M1, 32, E8M0> — a block-scaled
│ vector has no arithmetic in the OCP MX spec (only the block dot product).
× cannot index a `ScaledVec<FP4E2M1, 32, E8M0>` — a block is one packed value,
│ not an array. Use `scaled_dequantize(b)[i]` to read element `i` as FP32.
```

An element read only means anything as `X * P_i`, so it goes through `scaled_dequantize`; handing back the raw code without its scale would be a numeric bug with no diagnostic.

## `split` is a boundary shape, and needed glue

`split` emits `c_scale` + `c_elems` instead of one packed word, flattened the way bus ports already flatten — a real datapath feeds the scale into the exponent path and the elements into the mantissa path, so packing a 136-bit word per block is pure overhead.

It changes **no ARCH-level semantics**: the body still refers to the block by its single name, with compiler-emitted glue.

```systemverilog
logic [135:0] s;
assign s = {s_scale, s_elems};
assign s_out_elems = s_out[127:0];
assign s_out_scale = s_out[128+:8];
```

My first version omitted that glue and emitted `assign s_out = s;` against names that no longer existed — valid-*looking* SV that no frontend would accept. The legality check also had to move from the parser to typecheck, because type aliases resolve after parsing (`split MXFP4` is still `Named` at parse time). `arch sim` has no SV boundary, so it keeps the packed member.

## Package-scope type aliases

`type Name = TypeExpr;` is now legal in a `package` body and publishes file-wide, exactly like a package's structs and enums. The named formats are shared vocabulary, and module-scope-only aliases meant every module had to redeclare `MXFP4`. This reuses existing package semantics (`resolve.rs` already registers package contents as globals) rather than inventing file-scope aliases — no new scoping rule, no `use` plumbing.

A module alias reusing a package alias name is an **error**, not a silent shadow: package contents are global, so the collision is genuine ambiguity.

This is new surface beyond the §9 decisions, so it lands as its own spec section (§3.12) rather than buried in the ScaledVec text.

## A pre-existing bug this tripped over

Two `_ => 32` width fallbacks in `sim_codegen`. `type_bits_te_with_params` enumerated float formats **by hand**, so FP6E2M3 / FP6E3M2 / E8M0 have been silently reporting **32 bits** since #842/#848 added them to the format table but not to that list — no compile error to catch it (part of #829). Both now consult the canonical table.

This was not cosmetic: with the wrong width a 136-bit block took the *scalar* trace path and emitted C++ doing `(VlWide >> i) & 1`, which does not compile. `arch sim` was broken for the type until this was fixed.

## Verification

- `cargo test` green (fp suite 57 → 59)
- **Verilator 5.048 `-Wall` clean** on both emitted modules
- **Native sim round-trip 4/4**: 136-bit and 264-bit blocks survive as single packed values; the scale field is independent of the element plane
- Both rejections fire with spans
- A second module consumes the package alias with **no redeclaration** (`MXFP6` → `[199:0]`)

Spec: §3.11 ScaledVec, §3.12 package-scope aliases, type-table row.


nogate-ai:
## Self-review findings + fixes (pushed as 5c5b8b82)

Reviewed this PR adversarially and found **two blocking issues**, both now fixed.

### 1. `Elem`/`Scale` unvalidated → a real SV/sim divergence

All of these type-checked clean:

| Written | SV | sim |
|---|---|---|
| `ScaledVec<FP32, 32, E8M0>` | `[1031:0]` | — |
| `ScaledVec<FP4E2M1, 32, FP32>` | `[159:0]` | — |
| **`ScaledVec<UInt<8>, 4, E8M0>`** | **`logic [39:0]`** | **`uint8_t`** |

The third is a miscompile: a 40-bit port against an 8-bit variable. The SV emitter had its own recursive width walk that computed `8 + 8*4`, while the sim used the strict shared helper, got `None`, and `unwrap_or(0)`'d — a silent default, in a PR whose premise is that silent defaults are the bug class.

Fixed at the source with `is_block_element`/`is_block_scale` in the format table, enforced at every declaration and consulted by **both** width paths so they cannot disagree again. `FP8E4M3` is explicitly rejected as a scale — NVFP4's `UE4M3` is a different format, so substituting ours would be a real bug. `N = 0` is rejected too.

### 2. `==` rejected → inconsistent, and it made this PR's formal support unreachable

`Vec == Vec` and `struct == struct` already lower to a plain SV `==`. The blanket ban meant no formal property or TB check could mention a block, even though `check_scalar_type` accepts `ScaledVec`.

`==`/`!=` are now allowed as an **encoding compare**. Equal bits always mean equal values; unequal bits do *not* mean unequal values (same numbers, different scale/element split; and a NaN block ignores its element bits on load per decision #4). Sound but not complete as a value test — use `scaled_dequantize` for that. Arithmetic and **ordered** compares stay rejected: an encoding order is not a value order.

The property that previously could not be written now proves:
```
[Assert] passthru    PROVED  — up to bound 3
```

### Verified working (no change needed)

`split` + `port reg` emits correctly and lints clean; `.archi` round-trip and separate compilation; package/module alias collision errors; alias cycles caught when used; `==` exercises all 136 bits in both backends including the top word.

### Non-blocking, not fixed

- Package aliases permit forward references while module aliases document "no forward references" — package is more permissive; undocumented.
- An *unused* circular alias is silently accepted (cycle reports only on use) — matches existing module behavior.

`cargo test` green (fp suite 57 → 60), Verilator `-Wall` clean.

Issue #861, opened 2026-08-10
SILENT MISCOMPILE: non-atomic index base referencing a for-loop variable drops its parens (`(a + v[i])[3]` → `a + v[i][3]`)


nogate-ai:
## Summary

When an `Index` base is non-atomic **and** references a live runtime `for`-loop variable, the hoist bails and the emitter falls back to raw text concatenation of the base and `[idx]` — **without parentheses**. The result is the exact precedence miscompile arch#650 was filed to fix, reintroduced for the loop-var case.

Verilator accepts the wrong output silently. This is a wrong-values bug, not a portability bug.

## Repro (silent — Verilator compiles it clean)

```arch
module LoopSilent
  port a: in UInt<8>;
  port v: in Vec<UInt<8>, 4>;
  port y: out Vec<Bool, 4>;
  wire w: Vec<Bool, 4>;
  comb
    for i in 0..3
      w[i] = (a + v[i])[3];
    end for
    y = w;
  end comb
end module LoopSilent
```

Intent: `w[i]` = bit 3 of `(a + v[i])`.

Emitted:

```systemverilog
      w[i] = a + v[i][3];
```

which SystemVerilog parses as `a + (v[i][3])` — add bit 3 of `v[i]` to `a`, then truncate to 1 bit. Completely different value.

```
$ verilator --lint-only -Wno-fatal LoopSilent.sv
(0 errors — compiles clean)
```

iverilog rejects this particular shape, but only because of its own unrelated restriction ("Array index expressions must be constant here"), so it is not a reliable backstop:

```
$ iverilog -g2012 -o /dev/null LoopSilent.sv
LoopSilent.sv:10: error: A reference to a wire or reg (`i') is not allowed in a constant expression.
```

## Loud variant (same root cause)

```arch
      w[i] = (v[i] - 1)[3];
```
emits `w[i] = v[i] - 1[3];` → both frontends reject (`1[3]` is a syntax error). Same dropped-parens defect; it just happens to fail loudly because of the literal.

## Root cause

`src/codegen/mod.rs`, the `ExprKind::Index` arm. The hoist is skipped when the base references a live runtime `for`-loop variable — a module-scope temp cannot see a loop-local iterator — and the bail path does:

```rust
let b = self.emit_expr_str(unwrapped);
let i = self.emit_expr_str(idx);
return format!("{b}[{i}]");
```

[...]

Issue #827, opened 2026-08-09
Enhancement: multi-frontend SV portability sweep + baseline — portability is asserted, never measured (found 49 designs, 3 new bugs)


nogate-ai:
## Summary

ARCH's claim to emit portable SystemVerilog is currently **asserted, never measured**. The only SV frontend any CI job ever runs is Verilator — and no CI job runs it on `arch build` output at all. The compiler's internal notion of "portable" is a hand-maintained allowlist (`is_portable_bit_slice_base`, src/typecheck.rs:4554) whose own doc comment claims "Verilator/iverilog accept" these forms, a claim that has never been checked against iverilog and is currently wrong in both directions (#810, #813).

The result is a bug class that arrives one issue at a time, always discovered by a human happening to run iverilog: #650 → #653 → #807 → #808 → #810 → #813, all inside five weeks.

This proposes replacing the assertion with a measurement: **a multi-frontend portability sweep over the whole regression corpus, with a checked-in baseline, wired as a nightly CI tier.**

To show it works, I built and ran it. It found **49 designs whose emitted SV no CI job would ever have compiled**, including three previously unreported bugs.

## Measurement

Freshly built binary from a clean `origin/main` worktree (`fabe42f1`, v0.71.0). Tools: **Icarus Verilog 12.0 (stable, v12_0)** and **Verilator 5.048**. Per file:

```bash
arch build "$f" -o "$sv" --no-auto-asserts
iverilog -g2012 -gsupported-assertions -t null "$sv"     # parse/elab only
verilator --lint-only -Wno-fatal --timing "$sv"          # only on iverilog failures
```

Corpus: all 892 `.arch` files under `tests/` and `examples/`.

| Outcome | Count |
|---|---|
| `arch build` failed (negative fixtures: RDC/CDC violation tests, standalone duplicate-defs) | 96 |
| Emitted SV, package/bus-only — no module to elaborate (n/a) | 32 |
| **Emitted a real module** | **764** |
| ├─ Icarus-clean | 715 |
| ├─ **Verilator accepts, Icarus rejects** | **41  (5.4%)** |
| └─ **Rejected by _both_ Verilator and Icarus** | **8  (1.0%)** |

[...]

PR #885, opened 2026-08-11
fix(sweep): make portability verdicts reproducible; keep it nightly-only (#813 P3)

nogate-ai:
## Summary

Implements **#813 P3** (and the mechanism #812 asked for): SV portability now gates on **pull requests**, not only nightly.

`nightly-portability.yml` (#827) already sweeps the whole corpus through Icarus *and* Verilator — so the expensive part was already built. What was missing is **tier**: nightly means a portability regression is found up to 24 hours *after* it merges. Every fix in the #650 → #807 → #810 → #845 → #846 → #852 → #861 chain emitted SystemVerilog that no CI job compiled at merge time, and each was caught by hand days or weeks later.

New workflow `sv-portability-pr.yml` runs the same `scripts/sv_portability_sweep.sh` on PRs touching `src/`, `tests/`, `examples/`, or the sweep's own inputs — **Verilator only**.

## Why Verilator only, and why no script change

Verilator installs from apt in seconds and is the project's primary simulator. Alone it would have caught, pre-merge:

- #810's size-cast form `8'(x)[hi:lo]` (Verilator rejects it too — that was the correction that raised #810's severity)
- the `8'hff[s +: 2]` literal case found while implementing #813 P1
- #861's emitted shape

Icarus stays in the nightly tier deliberately: it is slower, more version-sensitive, and the whole #807/#810 family is Icarus-only, so a tool bump there should not block merges.

**No sweep-script flag was needed.** The script already records a frontend whose binary is absent as `SKIPPED` and excludes it from the baseline diff, so on a Verilator-only runner the Icarus half simply doesn't gate and cannot produce false failures. Verified locally by running the sweep with `iverilog` off `PATH`: `regressions=0 improvements=0 changed=0 skipped=785 new=0`, exit 0.

Kept as a **separate workflow** rather than a job in `test.yml`, whose header documents a deliberate policy of installing no external SV frontends. That policy is sound for the required check; this job can be adopted into branch protection on its own terms.

## Two baseline-artifact fixes found while wiring it up

1. **`tests/portability_baseline.tsv` had drifted behind the corpus.** Four fixtures were present but unblessed — `UniversalSliceHoist` (from #875), `sim_fp_narrow_mx_regression`, `sim_wide_runtime_slice_u128`, `sim_wide_runtime_slice_vlwide`. Unblessed rows are reported as `new` and **never fail**, so they were silently unmonitored. All four now recorded `OK` on both frontends.

2. **A `BUILD_FAIL` `reason` embedded the per-run `mktemp -d` scratch path**, so every `--bless` rewrote that row with a machine-specific absolute path. The sweep now rewrites it back to the repo-relative form. Verified: a second `--bless` is now a byte-identical no-op. Note the verdict diff compares column 3 only, so this could never fabricate a REGRESSION — it was reproducibility of the checked-in artifact, not gate correctness.

## Scope — what this does *not* close

Stating this plainly because the issue framing ("the CI gap that let all six ship") is only half right, and I'd rather the limitation be recorded than discovered later:

- The sweep is a **parse/elaborate check**. It cannot catch a wrong-values miscompile. #861 was exactly that — the bad SV compiled cleanly on Verilator and simply computed something else.
- It only ever sees **expression shapes some fixture already contains**. None of the six bugs' shapes were in the corpus when they shipped, so this job would not have flagged them at the time. They are covered now, via the regression fixture each fix added.

So this closes the *timing* gap. Systematic base-kind × select-form coverage — the thing that would have caught them proactively — remains open on #813.

## Verification

- Verilator-only sweep (the exact PR-gate path, `iverilog` hidden): clean, exit 0, 45 s locally at `-j 8`.
- Full dual-frontend sweep after re-blessing: `regressions=0 improvements=0 changed=0 skipped=0 new=0`.
- `--bless` idempotency: second run byte-identical.
- Both workflow YAMLs parse; `bash -n` clean on the modified script.
- Full `cargo test --release --no-fail-fast`: no failures. (The 2 `formal_test` Lean-replay tests fail in any fresh worktree because `proofs/lean_thread_lowering/.lake` is a gitignored build artifact; 29/29 with it present.)

Docs: dated `doc/COMPILER_STATUS.md` entry, skill snapshots re-synced. No `Cargo.toml` bump.

Implements #813 P3.


nogate-ai:
## Rebased onto #889; sweep is now clean

Rebased onto `9b6663aa` (#889, the `arch build` idempotency fix that this PR's diff change surfaced). Status-doc conflict resolved keeping both entries.

**The Verilator-only path — exactly what the new gate runs — is now fully clean:**

```
SUMMARY  regressions=0 improvements=0 changed=0 skipped=787 new=0 removed=0
sv-portability-sweep: no drift vs tests/portability_baseline.tsv
```

### Added since the first push

**A third baseline-artifact fix, and the most important one: the diff had a hole.** It keyed on `(file, frontend)` and let `new`/`removed` pass without failing:

```awk
exit (regress > 0 || improve > 0 || changed > 0) ? 1 : 0
```

A design that builds emits one row per frontend; one that fails to build emits a single `file<TAB>-<TAB>BUILD_FAIL` row. So an `OK -> BUILD_FAIL` flip made the per-frontend keys vanish (`removed`) and one new key appear (`new`) — and passed **green**. `arch build` breaking a corpus design was invisible to this sweep in *every* tier, nightly included.

A vanished key whose file is still swept is now REGRESSION (baseline `OK`) or CHANGED; a key whose file genuinely left the corpus is still just reported. Verified with a synthetic case: `OK -> BUILD_FAIL` yields two REGRESSION lines and exit 1, while a deleted file stays benign.

The very first run of the corrected logic found two real bugs:

- **#887** — `arch build` was not idempotent (`duplicate definition: <Bus>` on rebuild). Fixed in #889, merged.
- **#892** — `arch build` emits `~~x` for a double bitwise-NOT, which Icarus 12.0 rejects (`~(~x)` is fine). Recorded honestly in the baseline as `iverilog FAIL` so the gate doesn't flag it as fresh drift; it will report as an IMPROVEMENT when fixed.

### Baseline changes in this PR

| row | change |
|---|---|
| `tests/fp_v1/ScaledVecType.arch` | new, `OK` both frontends |
| `tests/sim_bitnot_compound_regression.arch` | new, `iverilog FAIL` (#892) / `verilator OK` |
| `tests/icarus_portability/UniversalSliceHoist.arch` + 3 sim fixtures | were present but unblessed, i.e. silently unmonitored |
| `examples/nic400/Nic400System.arch` | `8 errors` -> `7 errors` — a real effect of #889 |

### Path-strip fix, corrected

My first attempt substituted `"$OUT_DIR"` literally and silently missed: on macOS `/tmp` is a symlink to `/private/tmp`, so `--out-dir /tmp/x` comes back from the compiler as `/private/tmp/x/...`. It now matches by shape (`<anything>/corpus/`). Verified: `--bless` twice in a row is byte-identical.

nogate-ai:
## Scope changed: nightly-only, no PR gate

Per maintainer decision, the PR-gate workflow is **removed** from this PR. `nightly-portability.yml` stays the only tier that measures the portability claim; its timeout goes 30 -> 60 min to absorb the prepass below.

The PR-gate variant was built and measured before being dropped, and that is worth recording rather than deleting silently: it worked and ran in ~2 min — *until* the determinism fix, which costs ~5x and would have put it at ~10-15 min on every `src/`-touching PR. Too expensive for the merge path.

## What this PR is now

The gate was never the valuable part. Building it exposed **five defects in the sweep itself**, and those are what remains here:

1. **Verdicts were not order-stable.** `arch build` resolves a dependency from a sibling `.archi` an earlier build emitted, so in one shared corpus tree a design scored differently depending on how many of its dependencies were built first. `tests/aes/aes_key_expand_mod.arch` measured `BUILD_FAIL` at `-j 8` locally and *built* on CI at `-j 4`. Proven directly: cold build reports `undefined module: AesSbox`; build `aes_sbox.arch` first and it gets one dependency further, failing on `AesRcon`. Fixed with an `.archi` prepass run to convergence (4 -> 907 -> 908 -> 909) before measuring. **Verified: identical verdicts at `-j 8` and `-j 2`** — precisely what diverged before.

2. **The corpus copy inherited the working tree's gitignored `.archi` files** — 153 in the tree this was found on, versus 4 committed in a clean CI checkout. Every local run, including my earlier blesses, measured different inputs than CI. Untracked `.archi` are now dropped from the copy, and the baseline was re-blessed from a clean tree.

3. **The diff had a hole.** It keyed on `(file, frontend)` and let `new`/`removed` pass without failing:
   ```awk
   exit (regress > 0 || improve > 0 || changed > 0) ? 1 : 0
   ```
   An `OK -> BUILD_FAIL` flip makes the per-frontend rows vanish and one `file<TAB>-<TAB>BUILD_FAIL` row appear, so **`arch build` breaking a corpus design exited 0** — invisible in every tier, nightly included. Now REGRESSION (baseline `OK`), IMPROVEMENT (the file now builds) or CHANGED; a file that genuinely left the corpus is still just reported.

4. **A `BUILD_FAIL` reason embedded the per-run scratch path**, so every `--bless` rewrote that row with a machine-specific string. Matched by shape now rather than by substituting `"$OUT_DIR"` — on macOS `/tmp` resolves to `/private/tmp`, so the literal never matched (my first attempt at this silently did nothing).

5. **The baseline had drifted behind the corpus** — several fixtures present but unblessed, and unblessed rows report as `new` and never fail, i.e. silently unmonitored.

## Payoff

The first run of the corrected diff surfaced two real bugs the sweep had been structurally unable to report:

- **#887** — `arch build` was not idempotent. Fixed in #889, merged.
- **#892** — `arch build` emits `~~x`, which Icarus rejects. Fixed in #893, open.

## Verification

- Full `cargo test --release --no-fail-fast`: **0 failures** across every binary; `cargo fmt --check` clean.
- Verdict determinism: `-j 8` vs `-j 2` byte-identical.
- `--bless` idempotency: second run byte-identical.
- Nightly workflow YAML parses; timeout 60 min; no `sv-portability-pr.yml` remains.

Baseline diff is 9 rows and each was reviewed: two new `ScaledQuant` rows, `aes_key_expand_mod` and `clk_div_counter` moving from spurious `BUILD_FAIL` to real per-frontend `MODMISSING` verdicts, `buf_mgr_sm` likewise (`iverilog FAIL` / `verilator OK`), and `buf_mgr`'s reason going 7 -> 2 errors.

Issue #905, opened 2026-08-11
ScaledVec Phase 5b: NVFP4 blocks (UE4M3 scale + two-level per-tensor scaling)


nogate-ai:
Second half of phase 5. The `UE4M3` scalar type ships separately (conversions, `is_nan`, all three backends, characterized SMT miters); this is making it usable as a **block scale** and adding NVFP4's second scaling level.

```arch
type NVFP4 = ScaledVec<FP4E2M1, 16, UE4M3>;   // + a per-tensor FP32 scale
```

Typecheck currently refuses a `UE4M3` scale with a message pointing here.

## Why this is not just widening a predicate

Every MX block op assumes the scale is a **power of two**, and UE4M3's is not (it has 3 mantissa bits). Concretely:

- **`scaled_quantize` cannot use its reciprocal trick.** The E8M0 lowering computes `inv = arch_e8m0_to_f32(254 - code)` — exact, because negating a power-of-two exponent is exact. There is no such identity for UE4M3, so dividing by the scale needs either a real FP32 divide (no `arch_f32_div` exists; float `/` is rejected in v1, and division equivalence is SAT-hard) or a reciprocal-and-multiply with its own error analysis.
- **The single-rounding property is lost.** Spec §3.11.1 documents that scaling by `2^-(code-127)` is exact so each element rounds exactly once. Under UE4M3 the division rounds too, so a quantized element is doubly rounded — which changes the error bounds in §3.11.3 and needs its own gappa analysis.
- **`scaled_dot`'s scale application changes.** `(Σ ⊗ X^A) ⊗ X^B` is currently exact absent overflow because both scales are powers of two (§3.11.2). With UE4M3 both multiplies round, so the ordering argument has to be redone.
- `fp_block::BlockScale` needs a `Ue4m3` variant; the enum exists precisely to force these matches to be revisited.

## Two-level scaling

NVFP4 additionally carries a per-tensor FP32 scale above the per-block UE4M3 one. That is a surface question (where does the tensor scale live — a module port? a param?) as much as a numeric one, and should be designed before it is built.

## Gate

Proposal §8 phase 5: "vs TransformerEngine vectors".

nogate-ai:
## Review before starting: the division question is settled, and one claim above is wrong

### 1. Reciprocal-multiply is NOT a safe substitute for division — and it fails exactly where NVFP4 data lives

[...]

PR #918, opened 2026-08-12
refactor(fp): route every block-scale decision through BlockScale

nogate-ai:
## Problem

`BlockScale` existed as an enum, but the lowerings named E8M0's scale helpers as string literals (`arch_f32_to_e8m0`, `arch_e8m0_to_f32`) and hardcoded its magic codes. The enum **documented an intent it did not enforce**: adding a second variant compiled with ZERO errors and would have emitted E8M0 scale decoding for an NVFP4 block.

No diagnostic, and both backends wrong *identically* — so the SV↔sim differential gate would not have caught it either. That is arch#904's exact shape.

## Fix

Every genuinely scale-specific decision is now an exhaustive `match` on `BlockScale`, so a second variant has to answer each one:

`tag` · `width` · `widen_fn` · `narrow_fn` · `nan_code` · `zero_block_code` · `max_finite_code` · `sv_reciprocal` · `cpp_reciprocal`

Each carries the reasoning for why UE4M3's answer *differs* rather than inherits:

- **NaN code** — `0x7F`, not `0xFF`, which would additionally set the padding bit the format requires to be zero.
- **All-zero-block code** — E8M0 has no zero encoding, so its `0x00` is the *minimum scale* (2^-127); UE4M3's `0x00` **is** zero.
- **Reciprocal** — exact here only because negating a power-of-two exponent is exact. No analogue for a mantissa-bearing scale: arch#905 measured 4.76% of tie-aligned inputs changing quantized code under `v * fl(1/X)` vs `fl(v/X)`.

## The guard test

A source-level test pins the helper names inside `impl BlockScale`. It is deliberately a **source** check, not an output check — stripping the enum-supplied names from emitted text and hunting for leftovers is vacuous, because the stripped string *is* the literal a leak would use. With one variant, only the source can distinguish "came from the enum" from "hardcoded the same characters".

## Verification — both halves mutation-tested, not assumed

| Mutation | Result |
|---|---|
| Hardcode `arch_f32_to_e8m0` back into `sv_quantize` | Guard test **FAILS** with exact `src/fp_block.rs:504` + offending source line |
| Add a `Ue4m3` variant to `BlockScale` | Exactly **nine** `E0004` errors — one per decision above (lines 78/83/90/97/104/113/121/136/143) |

Fast gate: `cargo build --release` green · `cargo test --release` green (818 + all integration suites, 0 failures) · `cargo fmt --check` clean · `cargo clippy` 0 errors, 0 new warnings in `fp_block.rs`.

Refactor only — no emitted output changes intended; `scripts/refactor_diff.sh` running as the long verification.

Prerequisite for NVFP4 phase 5b (#905).

### DIFF
--- a/src/fp_block.rs
+++ b/src/fp_block.rs
@@ -44,9 +44,29 @@ use crate::fp_format::{by_type_expr, FpFormatId};
 use std::fmt::Write as _;
 
 /// The scale format of a block. One variant today; `UE4M3` (NVFP4) joins it
-/// in phase 5. An enum rather than a bool so that addition forces every
-/// `match` here to be revisited — the shared-exponent arithmetic below is
-/// specific to E8M0's "same bias as FP32, no zero, `0xFF` is NaN" shape.
+/// in phase 5b (arch#905).
+///
+/// **Everything the lowering knows about a scale lives behind this enum.**
+/// That is not decoration — it is the mechanism that makes adding a variant
+/// safe. The lowerings below previously named `arch_f32_to_e8m0` and friends
+/// as string literals, so adding `Ue4m3` compiled with **zero errors** and
+/// would have emitted E8M0 scale decoding for an NVFP4 block: no diagnostic,
+/// no compile error, and both backends wrong identically, so the SV↔sim
+/// differential gate would not have caught it either (the same failure mode
+/// as arch#904). Every method here is an exhaustive `match`, so a second
+/// variant now fails to compile at each genuinely scale-specific decision.
+///
+/// The decisions that are E8M0-specific — and therefore each need a real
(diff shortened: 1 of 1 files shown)


nogate-ai:
**Long verification complete — `scripts/refactor_diff.sh origin/main`:**

```
refactor_diff: compared 1681 common file(s); 0 byte-differ; 0 base-only; 0 head-only.
refactor_diff: PASS — zero diffs between origin/main and HEAD (2b98883)
```

Confirms the "no emitted output changes" claim across the full regression corpus + backend_equiv fixtures — every `.sv`/`.archi`/sim `.h`/`.cpp` is byte-identical. All 9 CI checks green.

Issue #375, opened 2026-05-18
arch check does not enforce single-driver: multiple writes to one reg/wire pass silently


ChienTeLi:
# `arch check` doesn't enforce single-driver: multiple assignments to the same `reg`/`wire` are silently accepted, producing multi-driver SV

## Versions affected

- `v0.48.1` (latest tagged release)
- `main` HEAD (tested at `d0b5834`)

In both versions `arch check` reports `OK: no errors` for a range of cases
where a single signal has more than one driver. `arch build` then emits SV
with duplicate `assign` statements, multiple `always_ff` procedures, or
multiple sub-module outputs wired to the same net.

## What the checker currently catches vs misses

Tested 10 patterns. The checker enforces three rules:

| ✅ Caught — `arch check` rejects | What it covers |
| --- | --- |
| **C4** Comb `if` with no `else` and no default | implicit latch |
| **C5** Output port never assigned | undriven signal |
| **C6** `comb` block using `=` to write a `reg` | wrong block kind for storage class |

It does **not** flag the following — these produce SV with duplicate drivers:

| ⚠️ Missed — `arch check` passes | What slips through |
| --- | --- |
| **C1** Same `comb` block, two unconditional assigns to one wire | `assign w = …; assign w = …;` |
| **C2** Two separate `comb` blocks writing the same wire | same as above, different blocks |
| **C3** Same `seq` block, two unconditional `<=` to one reg | duplicate `<=` in one always_ff |
| **C7** Two `inst` outputs wired (`->`) to the same internal wire | two child-module outputs drive one net |
| **C-seq** Two separate `seq on clk rising` blocks writing the same reg | two `always_ff` driving one reg |
| **C-thread** Two `tlm_method` thread bodies writing the same reg | two `always_ff` (one per thread) driving one reg |

For C9 (default + override pattern) and C10 (two `if` in one `seq` writing the
same reg) the checker passes and the emitted SV is a single priority-mux
`always_ff` — those are not multi-driver and are correctly accepted.

## Repro A — plain `seq` × 2 (no TLM involved)

`test_two_seq_same_reg.arch`:

```arch
domain SysDomain
  freq_mhz: 100
end domain SysDomain

module TwoSeqSameReg
  port clk: in Clock<SysDomain>;
  port rst: in Reset<Sync>;
  port enq_en: in Bool;
  port deq_en: in Bool;
  port valid:  out Bool;

  reg valid_r: Bool reset rst => false;

  comb
[...]

PR #924, opened 2026-08-12
feat(inst): add `auto;` auto-connect directive

### LINKED ISSUE #754: Enhancement: `auto;` — explicit auto-connect for name-matching `inst` ports
## Summary

Add an explicit **auto-connect directive** inside `inst` blocks: a single `auto;` line that wires every still-unconnected sub-module port to the **identically-named, in-scope, direction/type-compatible** signal. Explicit `port <- sig` / `port -> sig` connections still take precedence; `auto;` only fills the remainder.

```arch
inst alu0: Alu
  a <- Decode.rs1_fwd;    // explicit where names differ
  b <- Decode.rs2_val;
  auto;                   // clk<-clk, rst<-rst, en<-en, result->result, ...
end inst alu0
```

## Problem — measured conciseness gap #5 (multi-module decomposition overhead)

`inst` + N× `connect` is the worst-case length offender vs the SystemVerilog reference in the VerilogEval comparison: `Prob068_countbcd` (4 BCD counters) runs **205% of the SV reference**.

Evidence from the current example corpus (`examples/*.arch`, 2026-07-25): **13 of 34 connection lines (38%) are identical-name `port <- port` boilerplate** — `clk <- clk; rst <- rst; en <- en;`. Pure noise that costs tokens on every hierarchical design.

## Why it matters

ARCH is "explicitly designed to be generated correctly by LLMs." Shorter source = fewer tokens and fewer wrong-name typos. Clock/reset/enable plumbing is the most mechanical **and** most error-prone boilerplate — forgetting to wire `clk`/`rst` into a sub-instance is a classic bug. `auto;` makes the common correct case the default while the compiler still **errors** if a same-named signal is absent (all-ports-connected preserved) or its type/direction mismatches (no-implicit-conversion preserved). Boilerplate goes away; no safety invariant is weakened.

## Design — keeps ARCH's explicit/reviewable ethos (the differentiator from SV silent `.*`)

- **Syntax:** a dedicated `auto;` line inside the `inst` body (avoid overloading `.`; fits the keyword style). Bikeshed: `auto;` vs `connect matching;`.
- **Semantics:** for each sub-module port **not** already explicitly connected, resolve the in-scope signal (port/wire/reg/let) with the **identical name**; require compatible direction (in-port pulls from the same-named signal, out-port drives it) and exact type — reusing the existing connection type/direction checks (**no new semantics**). No match → error `auto-connect: no signal 'X' in scope for port 'X'`. Type/dir mismatch → the existing inst-connection type error. Names are unique in scope, so no ambiguity.
- **Reviewability:** `auto;` is a visible, greppable token; a `--explain-auto` verbose mode (or a compile note) prints the expanded connection list. Anything non-name-matching must still be written explicitly.

## Implementation approach — front-end desugar, low backend risk

1. **Parser** (`src/parser.rs`, inst-block parse): accept `auto;`, set a flag on the `Inst` AST node.
2. **Elaboration/typecheck:** after processing explicit connections, iterate the unconnected sub-module ports and synthesize `port <- sig` / `port -> sig` for each identical in-scope name, validated by the **existing** connection checker.
3. **Codegen unchanged:** connections are fully resolved before SV/sim/`.archi` emission, so all three backends see the same explicit connection set — no native<->SV divergence risk (pure desugar).
4. **Docs + fixtures:** update the spec inst/connect section + `doc/Arch_AI_Reference_Card.md`, add `tests/` and `examples/` fixtures.

## Relationship to other issues

Complementary to #737 and #738 (checking width/signedness/Vec-length and Clock/Reset conformance on inst connections): `auto;` synthesizes connections through the **same** connection checker, so it must respect — and benefits from — those conformance checks rather than bypassing them.

## Governance

This adds user-facing syntax (grammar-surface change), so per `CLAUDE.md` it needs maintainer sign-off and a spec update in the same PR before implementation.

nogate-ai:
Implements the `auto;` auto-connect directive from #754.

A single `auto;` line inside an `inst` body connects every child port the explicit connections left unconnected to the **identically-named** signal in the enclosing scope. Explicit connections always win, wherever they sit relative to the directive.

```arch
inst alu0: Alu
  a <- Decode.rs1_fwd;    // explicit where the names differ
  b <- Decode.rs2_val;
  auto;                   // clk <- clk, rst <- rst, en <- en, result -> result, ...
end inst alu0
```

## Why

13 of 34 connection lines in `examples/*.arch` (38%) are identical-name `port <- port` boilerplate. `inst` + N connections is also ARCH's worst length offender against the SV reference (`Prob068_countbcd` = 205%). Clock/reset plumbing is the most mechanical *and* most error-prone part of a hierarchical design — and today's unconnected-port check in `typecheck.rs` deliberately **skips** Clock and Reset ports, so a missing `clk` isn't even diagnosed. `auto;` makes the common correct case one line while keeping every fill visible (`--explain-auto`) and checked.

## How

A pure front-end desugar: the new `elaborate::expand_auto_connect` pass rewrites `auto;` into ordinary `Connection`s right after `elaborate()` — at that point inst-body `for` loops are flattened, generate blocks expanded, `connect` sugar appended, and module variants monomorphized (so child port widths resolve). Resolve, typecheck, and all three backends see an ordinary explicit connection set and know nothing about `auto;`.

The headline test asserts the property directly: **emitted SV is byte-identical to the hand-written twin**.

**Fills**: ordinary ports (incl. `Clock`, `Reset`, `Vec`, struct, enum), whole `bus` ports (incl. `Vec<Bus, N>`), and `ports[N]` groups by flattened name (`read0_addr`; `write_en` for a literal count-1 group, matching `normalize_count1_portarray_conns`).

**In scope**: the enclosing construct's ports, its `reg`/`wire`/`let`/`pipe_reg` declarations, flattened bus fields, and the targets of *explicit* inst output connections. Names created by another inst's `auto;` are excluded, so expansion is order-independent.

Works in module bodies, inside `generate_for` / `generate_if` (including the SV-genvar-preserved form), and in `pipeline` stage bodies.

## Safety — both invariants preserved, not weakened

- **All ports connected.** No name match is an error, for inputs *and* outputs. `auto;` never leaves a port dangling and never silently declares a net.
- **No implicit conversion.** Each fill is type-checked against the declared signal before synthesis. A definite mismatch — width, signedness, `Vec` length, reset kind/polarity, clock domain, named type — is an error naming both sides, with the `as Reset<...>` override suggested for the reset case. This matters because inst connections are otherwise **untyped** today (#737), so without the gate `auto;` could wire an inverted reset silently. The comparison is deliberately conservative: param-dependent widths and undeclared inst-output wires defer to the existing checks rather than inventing a type rule here.

`auto` is recognized **contextually** — only as a complete `auto;` statement in an inst body — so it stays a legal identifier everywhere else and a port genuinely named `auto` still connects (`auto <- en;`). Duplicate `auto;` and `auto;` inside an inst-body `for` loop are parse errors. A stale `auto;` that fills nothing warns.

Designs that don't use `auto;` are untouched: the pass early-returns when no directive is present in the file.

## Also in this PR (reuse for the new pass)

- `Item::ports()` / `Item::params()` / `Item::port_arrays()` in `ast.rs`, replacing the duplicated `item_ports` / `item_params` helpers in `main.rs` and `graph.rs`.
- `BusInfo::from_decl` extracted in `resolve.rs` so a pre-resolve pass can reach `effective_signals`.

## Docs

Grammar-surface change, so `doc/ARCH_HDL_Specification.md` (inst-connections section) and `doc/Arch_AI_Reference_Card.md` are updated here, plus the skill reference snapshots (`scripts/sync_skill_snapshots.sh refresh`). The reference card is what the MCP server serves as `get_construct_syntax("module")`, so agents pick this up automatically.

## Verification

- `cargo test` — full suite green (14 new `auto_connect` tests).
- New tests cover: SV byte-equality vs explicit, explicit-wins precedence, unmatched input/output errors, width and reset-polarity gate errors, param-dependent deferral, whole-bus fill, `ports[N]` + count-1 spelling, `generate_for`, `auto` as an identifier, duplicate/for-loop parse errors, and the no-op warning.
- Manually verified end-to-end with `arch check`/`arch build --explain-auto` on module, bus, regfile-`ports[N]`, `generate_for`, and pipeline-stage fixtures.
- `scripts/refactor_diff.sh origin/main` (zero-diff over the regression corpus) runs after this PR opens, per fast-gate-before-PR ordering.

closes #754

### DIFF
--- a/doc/ARCH_HDL_Specification.md
+++ b/doc/ARCH_HDL_Specification.md
@@ -2049,6 +2049,26 @@ The compiler validates that the default value fits within the declared range. Co
 >
 > ◈ **Port group member syntax.** For constructs with named port groups (e.g. `regfile` with `ports[N] read`), connections use dot notation: `write.en <- wr_en;` (flattened to `write_en`). For indexed port groups, use bracket-dot notation: `read[0].addr <- sel;` (flattened to `read0_addr`). The index must be an integer literal. Both forms are resolved at parse time --- the rest of the compiler sees only the flattened name.
 >
+> ◈ **`auto;` --- auto-connect by name.** A single `auto;` line inside an `inst` body connects every child port that the explicit connections left unconnected to the **identically-named** signal in the enclosing scope. Explicit connections always win, wherever they appear relative to the directive:
+>
+> ```
+> inst alu0: Alu
+>   a <- Decode.rs1_fwd;    // explicit where the names differ
+>   b <- Decode.rs2_val;
+>   auto;                   // clk <- clk, rst <- rst, en <- en, result -> result, ...
+> end inst alu0
+> ```
+>
+> `auto;` is a **front-end desugar**: elaboration rewrites it into ordinary connections before every later pass, so emitted SystemVerilog, simulation, and formal output are byte-identical to the hand-written connection list. Run `arch check --explain-auto` (also on `build` / `sim` / `formal`) to print exactly what each directive expanded to.
+>
+> **What it fills:** ordinary ports (including `Clock`, `Reset`, `Vec`, struct and enum types), whole `bus` ports (`initiator` / `target`, including `Vec<Bus, N>`), and `ports[N]` groups by their flattened names (`read0_addr`, and `write_en` for a literal count-1 group).
+>
+> **In scope** are the enclosing construct's ports, its `reg` / `wire` / `let` / `pipe_reg` declarations, the flattened fields of its bus ports and bus wires, and the targets of *explicit* inst output connections. Names created by another inst's `auto;` are deliberately **not** in scope, so expansion never depends on inst order.
+>
+> **Both safety invariants are preserved, not weakened.** A port with no identically-named signal in scope is a compile error (*"auto-connect: no signal \`a\` in scope for input port \`a\` of \`Alu\` in inst \`u0\`"*) --- `auto;` never leaves a port dangling, and never silently declares a new net. And before each fill the port type is compared against the declared signal type: a definite mismatch --- differing width, signedness, `Vec` length, reset kind or polarity, clock domain, or named type --- is an error naming both sides, with the `rst <- rst as Reset<Async, Low>` override suggested for the reset case. The comparison is deliberately conservative: when a width stays param-dependent, or the name comes from an inst-output wire with no declaration, `auto;` defers to the ordinary connection checks rather than inventing a type rule.
+>
+> `auto` is a **contextual** keyword recognized only as a complete `auto;` statement inside an inst body --- it remains a legal identifier everywhere else, so a port genuinely named `auto` still connects as `auto <- en;`. Two `auto;` lines in one inst body, or an `auto;` inside an inst-body `for` loop, are compile errors.
+>
 > ◈ **Whole-bus connections.** When an inst's bus port connects to a parent bus port (or wire), a single connection expands to all signals in the bus definition: `axi_rd -> m_axi_mm2s;` expands to `axi_rd_ar_valid -> m_axi_mm2s_ar_valid`, etc. Signal directions are derived from the bus definition and the port's perspective (`initiator` or `target`). This works for both `module` and `fsm` constructs.
 >
 > ◈ **Prototype TLM connect sugar.** For early design-phase TLM binding, a module may write `connect cpu.mem -> ram.mem;` after both instances are declared. The left endpoint must be an instantiated `initiator` bus port and each right endpoint an instantiated `target` bus port of the same bus type. Elaboration creates private bus wires and appends ordinary whole-bus inst connections, so generated SV remains the same flattened req/rsp signal protocol. `connect` is legal at module scope and inside `generate_for` / `generate_if`; generated suffix names are expanded first (`connect src_i.m -> dst_i.s;` becomes independent `src_0.m -> dst_0.s`, `src_1.m -> dst_1.s`, ... links). Blocking TLM buses also support one-initiator-to-many-target address decode by repeating ordinary connect statements and overriding each target instance's address-map params:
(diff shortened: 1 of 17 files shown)


nogate-ai:
Long verification complete (post-PR, per fast-gate-before-PR ordering):

- **`scripts/refactor_diff.sh origin/main`: PASS** — 1681 emitted files compared, **0 byte-differ, 0 base-only, 0 head-only**. The pass early-returns when a file contains no `auto;`, so it is provably inert on the whole existing corpus.
- **SV byte-equality against hand-written twins** for all three port shapes, using a freshly built binary: plain module ✅, `regfile` `ports[N]` groups (incl. the count-1 `write_en` spelling) ✅, whole-`bus` ✅. This is the property that matters — `auto;` emits exactly what the explicit connection list emits.
- **Verilator 5.048 lint** on the fixtures: no `PINMISSING`, `PINCONNECTEMPTY`, or `WIDTH` warnings — the failure modes an auto-connect expansion could introduce. Residual warnings are fixture properties (toy leaves that ignore some inputs; an unused `regfile` type param) and appear identically on the explicit twins, which are byte-identical.
- `cargo fmt --all --check` and `cargo clippy --all-targets` clean.

Unrelated finding while verifying, filed as #929 (**not** caused by this PR — the repro contains no `auto;`, and refactor_diff is zero-diff): `arch check` misses the single-driver rule when the instances come from a `generate_for`, a residual corner of the closed #375.

nogate-ai:
Daily review sweep (2026-08-22): flagged STRANDED (green checks, ≥1d old) but now **CONFLICTING/DIRTY** — `main` has moved since this was opened and it needs a rebase to resolve conflicts. This is a **grammar-surface** feature (`inst … auto;`), so the automated sweep will not autonomously rebase-and-merge a language-surface change in an unattended run. Labeling `awaiting-decision`: a maintainer should rebase onto current `origin/main`, re-run the fast gate, and merge. The four design decisions were previously user-approved (see memory: auto-connect inst / closes #754).

PR #926, opened 2026-08-12
feat(fp): UE4M3 as a ScaledVec block scale — NVFP4 (closes #905)

### LINKED ISSUE #905: ScaledVec Phase 5b: NVFP4 blocks (UE4M3 scale + two-level per-tensor scaling)

nogate-ai:
Phase 5b of the MX/NVFP4 line. `ScaledVec<FP4E2M1, 16, UE4M3>` now type-checks, quantizes, dequantizes and dots in both backends. Closes #905.

## The issue's premise turned out to be false

#905 predicted this needed an `arch_f32_div` and would cost the single-rounding property, because every MX block op assumes a power-of-two scale and UE4M3 carries three mantissa bits. **The quantizer never forms the quotient at all.**

Instead of dividing, it compares `|v|` against `X × m` for each RNE decision boundary `m` of the element grid. Every such product is **exact in FP32** — the scale has 4 significand bits and a boundary at most 5, so the product needs at most 9 of FP32's 24, with no exponent-range escape. That makes each comparison a decision rather than an approximation, and the selected code correctly rounded by construction.

Ties-to-even needs no runtime logic: exactly one of the two codes at a boundary is even, so it is a strict `>` at even `k` and a `>=` at odd `k` — fixed at generation time, per boundary. The scale is chosen the same way, against constant thresholds that fold entirely.

**So single-rounding is recovered, not lost, and the §3.11.3 error bounds carry over to UE4M3 unchanged.**

## Evidence

Over all 126 finite UE4M3 scales × all five element formats:

| Check | Result |
|---|---|
| `X × m` exact in FP32 (the property the kernel rests on) | **0 inexact** |
| Ladder vs a real divide — every boundary exactly, both 1-ULP neighbours, every grid point | **0 disagreements** |
| The reciprocal shortcut this replaces | **4.76% wrong** on tie-aligned inputs — reproduces the figure recorded in #908 |
| Emitted E4M3 overflow boundary | **464.0 with a strict `>`** — OCP's documented rule (464 ties down to 448, ≥480 overflows), arrived at from the parity rule rather than transcribed |

**Cross-backend (§8 gate):** `fp_nvfp4_sv_matches_sim` runs Verilator on the emitted SV and the native sim on the emitted C++ over 10 vectors — all-zero, signed-zero, NaN, Inf, subnormal, saturating, wide-range and E4M3-overflow blocks — and byte-compares the transcripts.

## Surface decisions (maintainer-approved 2026-08-12, recorded in `doc/proposal_mx_block_formats.md` §9.1)

- **The default scale policy is now per-scale**: `floor_pow2` for E8M0 (OCP §6.3, unchanged), `exact` for UE4M3. One global default cannot serve both — defaulting UE4M3 to `floor_pow2` discards all three mantissa bits and silently emits a power-of-two-scale block under the NVFP4 name. `exact` is *refused* for E8M0 rather than ignored. This amends proposal §9 decision #1.
- **NVFP4's per-tensor FP32 scale stays out of the type.** A `ScaledVec` is one block, so a fourth type parameter would replicate a per-tensor quantity per block. It costs nothing numerically: fusing it in would compare against `S × X × m`, which is *not* exact for arbitrary `S` — where `X × m` is — so the fused form would round where the split form does not.

## Latent bug found by the gate

The 6th instance of the arch#837 hazard: `block_member_width` knew `E8M0` but not `UE4M3`, so `scaled_vec_width` returned `None` and the sim's fallback sized a 72-bit block as a `uint8_t` while the SV emitter kept the full width. Exactly the phase-2a failure its own doc comment describes. Added `block_member_width_covers_every_legal_scale`, which fails on any future scale admitted by `is_block_scale` but not measurable.

## Verification

- `cargo test` green (2 pre-existing `construct_proof_lean_*` failures are local Lean-toolchain env issues — confirmed identical on clean `origin/main`).
- All new specs **mutation-tested, 4/4 mutants killed**: tie parity, the NaN code, the per-scale default, and the underflow clamp each fail a named test.

### DIFF
--- a/doc/ARCH_HDL_Specification.md
+++ b/doc/ARCH_HDL_Specification.md
@@ -838,9 +838,10 @@ to a scale, and `is_nan` tests the single code `0x7F`. Because a scale is
 non-negative, `.to_ue4m3()` takes the **magnitude** of its operand (as
 `.to_e8m0()` does) and always clears the padding bit.
 
-`UE4M3` is a scalar type today. Using it as a `ScaledVec` scale is not yet
-supported --- see §3.11 --- because the block operations depend on the scale
-being a power of two.
+`UE4M3` is also a `ScaledVec` scale --- that is what it exists for --- giving
+NVFP4 as `ScaledVec<FP4E2M1, 16, UE4M3>`. The block operations do not need a
+power-of-two scale: `scaled_quantize` never divides by the scale at all, so
+the single-rounding property survives. See §3.11.1.
 
 
 
@@ -1071,7 +1072,12 @@ ScaledVec<Elem, N, Scale>
 Microscaling (MX) formats. The scale is a **parameter, not a fixed type**,
 because the two ecosystems genuinely differ --- MX uses `E8M0`, NVFP4 uses a
 `UE4M3` that is *not* our `FP8E4M3` (unsigned, 7 significant bits, NaN
-`0x7F`). `UE4M3` and NVFP4 are not yet implemented.
+`0x7F`). Both scales are supported:
+
+```arch
+type MXFP4 = ScaledVec<FP4E2M1, 32, E8M0>;
+type NVFP4 = ScaledVec<FP4E2M1, 16, UE4M3>;
+```
 
 The three arguments are constrained:
 
@@ -1080,11 +1086,13 @@ The three arguments are constrained:
     scale has no meaning over wider or integer elements.
   - **`N`** is a const expression `>= 1`. A block with no elements is just a
     bare scale.
-  - **`Scale`** must be `E8M0`. `UE4M3` (§3.10a) exists as a scalar type but
-    is not yet accepted here: unlike `E8M0` its value is not a power of two,
-    so dividing by it is inexact and the block operations' single-rounding
-    property (§3.11.1) does not hold. `FP8E4M3` is **not** a stand-in for
-    either.
+  - **`Scale`** must be `E8M0` (OCP MX) or `UE4M3` (NVFP4, §3.10a).
+    `FP8E4M3` is **not** a stand-in for `UE4M3` --- that one is signed and
+    its NaN is sign-agnostic. The two scales differ in more than width:
+    `E8M0` is a pure power of two with no zero, `UE4M3` carries three
+    mantissa bits, *has* a zero at `0x00`, and reserves `0x7F` (not `0xFF`)
+    for NaN. Those differences reach the surface, so they are spelled out
+    per operation below rather than left implicit.
 
 **A block is one packed word, not an array.** Layout is
 `{ scale[w-1:0], P[N-1], …, P[1], P[0] }` --- scale in the high bits,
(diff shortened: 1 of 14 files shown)


nogate-ai:
**Long verification complete** (per fast-gate-before-PR ordering — the fast gate ran before this PR opened; these are the long checks):

**`scripts/refactor_diff.sh origin/main` → PASS: 1681 files compared, 0 byte-differ, 0 base-only, 0 head-only.**

That is the check this PR most needed. It restructures code the `E8M0` path shares — `sv_reciprocal`/`cpp_reciprocal` became `Option`, and `sv_quantize`/`cpp_quantize` gained a kernel dispatch — so the existing MX lowering had every opportunity to shift. It emits byte-identical SystemVerilog, `.archi` and sim C++ across the whole regression corpus. The new UE4M3 path is purely additive.

Also: `cargo test` green apart from two `construct_proof_lean_*` failures, which are a local Lean-toolchain environment issue — **confirmed failing identically on a clean `origin/main` worktree**, so they are not from this branch.

Pushed `2e46adf4` for the one `cargo fmt` hunk (a test assertion) that CI flagged.

PR #949, opened 2026-08-22
fix(typecheck): reject undefined value identifiers (#748)

### LINKED ISSUE #748: arch check accepts undefined value identifiers — emits SV with undeclared nets, reports "OK: no errors"

nogate-ai:
## What

`arch check` accepted references to identifiers that are **not declared anywhere** — no port, `wire`, `reg`, `let`, or `param`. It printed `OK: no errors` and exited 0; `arch build` then emitted SystemVerilog referencing undeclared nets, which no frontend accepts.

```arch
comb
  o2 = missing_one;   # arch check: "OK: no errors", exit 0
end comb
```

```
%Error: undef.sv:7:15: Can't find definition of variable: 'missing_one'
```

`arch check` is the fast pre-flight gate before `build`/`sim`, so a clean check on a design that cannot possibly build inverts its contract — and a hallucinated signal name is exactly the failure mode an LLM-targeted language most needs it to catch.

## Why it was silent

The `ExprKind::Ident` arm of `resolve_expr_type` fell through to `Ty::Error` for any name missing from `local_types`. `Ty::Error` is the poison value that suppresses downstream diagnostics, so not only was nothing reported here, the surrounding expression's checks went quiet too.

That `else` was doing double duty: it was also the fallback for legal-but-generically-typed names (`param` references, enum variants). Blanket-erroring on it would have broken those, so the fix resolves the name explicitly and errors only when **every** lookup misses.

## Resolvable surfaces

Each was confirmed legal against the regression corpus before being added:

| Surface | Example |
|---|---|
| `param` in the enclosing construct | `o = a +% SCALE;` |
| Names in the construct's own symbol scope | ports / regs / lets / wires |
| File- and package-level globals | enum & struct type names, functions |
| Bare enum variants | `c = Red;` — not a global in its own right; lives in its enum's variant list |
| Nets implicitly declared by an `inst` **output** connection | `clk_out -> clk_ifu_w;` — SV codegen emits a `logic` for these |
| `for` loop iterators | the caller never inserted these into `local_types`; the loop body now binds its own |
| The `__`-prefixed compiler-reserved namespace | `__chwire_*`, `__prop_*`, `__staged_*` — a user `assert` may reference one before the pass that materializes it has run |

Loop iterators are bound as **`Ty::Error` deliberately**: that is the type they effectively had before this change, so width and latency checks on iterator-derived expressions keep their previous behaviour and only name *resolution* changes. Giving iterators a real integer type is a worthwhile follow-up but is a separate, wider-blast-radius change.

## Verification

The risk here is entirely **false positives** — legal code the new check rejects. So the gate was a differential sweep of `arch check` against a binary built from the base commit, over the whole corpus, run two ways:

- **Per-directory** (68 dirs) — whole-design context, so cross-file `inst` references resolve.
- **Per-file** (933 files) — because directories holding independent fixtures (`tests/formal`) die on `duplicate definition` before typechecking and would otherwise get no coverage at all.

Final result on both: **zero exit-code changes, zero new diagnostics.**

Two of the seven resolvable surfaces above were *found by that sweep* rather than by reading code, which is the argument for running it:

- `tests/e203` went clean → 9 errors on inst-output-bound nets (`clk_out -> clk_ifu_w;`).
- `tests/formal` broke on `__chwire_*` lifted credit-channel state referenced from user `assert` bodies.

Fast gate: `cargo test --release --no-fail-fast` — **1225 passed, 0 failed**; `cargo fmt --check` and `cargo clippy --release --all-targets` clean.

## Tests

Five regression tests: the bug itself, all four value-expression contexts from the issue (`let` initializer, arithmetic sub-expression, `comb` if-condition, seq RHS), and one guard per must-not-break surface (param + bare enum variant, `for` iterator, inst-output net).

## Spec

Because this changes which programs the compiler accepts, `doc/ARCH_HDL_Specification.md` gains the rule in the §3.4 static-rules block, alongside the existing no-implicit-latches and single-driver rules. Skill snapshots re-synced via `scripts/sync_skill_snapshots.sh refresh`.

## Note

`arch check` on the repro now reports **both** undefined names in one run rather than the first only — that is #750's batch-diagnostics work already in `main`, not something added here.

closes #748


nogate-ai:
Long verification (post-open, per the repo's fast-gate-before-PR ordering):

**`scripts/refactor_diff.sh origin/main` — PASS.** Compared **1669** emitted files across the baseline regression corpus and the `backend_equiv` fixtures: `0 byte-differ; 0 base-only; 0 head-only`.

That is the property this change should have — it adds a diagnostic and must not move codegen. Every design that compiled before emits byte-identical SystemVerilog, `.archi`, and sim C++.

All CI checks green (`cargo test`, clippy, fmt, `examples/run_sims.sh`, doc drift, skill snapshots, release doc, tb registered, CLA).

PR #951, opened 2026-08-23
fix(codegen): make user-function bodies parseable by yosys

nogate-ai:
## Problem

Emitted `function automatic` bodies used two constructs yosys's built-in Verilog frontend rejects — both **legal SystemVerilog** that Verilator 5.048 and Icarus 12.0 accept, so no simulator gate in CI ever saw them:

```sv
logic [7:0] d = expr;   // "Invalid nesting of always blocks and/or initializations"
return expr;            // "syntax error, unexpected TOK_ID"
```

#950 fixed the first for the generated soft-float helper library. This fixes **both** for user-written `function` bodies.

Every ARCH function must reach a `return` (the no-latch rule), so the `return` gap alone made *every* design containing a user function unsynthesizable — and it **masked** the declaration gap, since parsing dies on `return` first. That's why the two must land together to be observable at all; the declaration fix on its own measures as zero change.

## Measured

14 corpus designs declare a user function containing a `let`. Yosys frontend parse:

| | pass | fail |
|---|---|---|
| before | **0** | 14 |
| after | **14** | 0 |

Combined with #950, this takes `tests/fp_v1` from **3/31 → 31/31**.

## Two changes

**1. Declarations hoist above the body**, leaving assignments in place. Deduplicated by name — which is what makes hoisting out of `if`/`for` arms safe. The compiler already models function locals as one flat namespace (`collect_fn_local_types` walks nested bodies into a single map), so a name reused across two arms stays one variable assigned twice, exactly as before, never two declarations.

**2. `return expr;` lowers to `<fn_name> = expr;`** — but only where no return is an early exit. The unsound shape:

```
if c        ->   if (c) begin
  return a;         fname = a;
end if            end
return b;         fname = b;   // early exit dropped: overwrites `a`
```

SV's `return` carries the control transfer; a bare assignment does not. Functions with that shape **keep emitting a literal `return`** — correct, and exactly today's behavior, merely not yosys-parseable. No regression, ever.

## Why the guard is trustworthy

An AST walk over the corpus — 900 files, 49 functions, 58 returns — finds **zero** early exits, so the guard never fires there. The language permits the shape though, so `early_return_keeps_literal_return` pins the fallback with a synthetic case that nothing else in the tree exercises.

Worth noting `sub_beats` in `Nic400Axi4ToAxi3` *looks* like an early return but is an `if`/`else` where both arms return — correctly classified as rewritable, with the `min(16, remaining)` clamp preserved.

## Test changes

Golden assertions that encoded the old emission now assert what they were actually protecting — the `$signed` wrapper, the `base + stepped` arithmetic, the `l == 1` lock mapping, the `min(16, remaining)` clamp — and no longer pin monomorphized names like `sub_addr_5_x_3`.

Two insta snapshots re-accepted after diff review: declarations hoisted, assignments in unchanged source order, every expression byte-identical.

## Verification

- `cargo test --release` — **1228 passed, 0 failed** (1225 baseline + 3 new tests)
- `cargo fmt --check` clean
- Verilator + Icarus unchanged across all 14 designs — 3 pre-existing failures on each side, identical before and after

Further progress on #932; its genuine-defect findings, including the broken reset in `examples/`, remain open.

### DIFF
--- a/src/codegen/mod.rs
+++ b/src/codegen/mod.rs
@@ -148,6 +148,9 @@ pub struct Codegen<'a> {
     /// `ident_decl_type`/`ident_float_fmt` so float-op dispatch works
     /// inside `function` bodies; empty outside `emit_function`.
     fn_local_types: std::collections::HashMap<String, TypeExpr>,
+    /// SV name to assign to in place of `return` inside the function being
+    /// emitted, or `None` to emit a literal `return` (see `emit_function`).
+    fn_return_target: Option<String>,
     /// Maps call-site span.start → overload index (for overloaded functions only).
     overload_map: std::collections::HashMap<usize, usize>,
     /// Bus port names in the current module → bus name (for FieldAccess rewriting).
@@ -419,6 +422,7 @@ impl<'a> Codegen<'a> {
             pending_functions: Vec::new(),
             file_scopes: Vec::new(),
             fn_local_types: std::collections::HashMap::new(),
+            fn_return_target: None,
             overload_map,
             bus_ports: std::collections::HashMap::new(),
             vec_of_bus_port_count: std::collections::HashMap::new(),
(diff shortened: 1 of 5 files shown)


nogate-ai:
**Long-verification numbers.**

Full `tests/fp_v1` corpus (31 designs), yosys frontend parse, built fresh from each side:

| | pass | fail |
|---|---|---|
| before #950 (`d56f86ee`) | 3 | 28 |
| after #950 (`792332ad`) | 30 | 1 |
| **with this PR** | **31** | **0** |

The whole float line — scalar FP, FP8/FP6/FP4, and the MX/NVFP4 block-scaled quantizers — now clears the yosys frontend, where before this pair of fixes only 3 of 31 did.

Separately, the 14 corpus designs declaring a user function with a `let` go **0/14 → 14/14**.

All 10 CI checks green.

Issue #955, opened 2026-08-24
Pipelined block operators: scaled_dot / scaled_quantize need registry rows (scaled_dequantize does not)


nogate-ai:
## Summary

The MX block operators are single-cycle combinational today. Measured against the one operator the registry already stages (`ArchF32FmaStaged6` — 6 stages, ~260 MHz), two of the three are firmly multi-cycle logic and need `<pipelined, N>` registry rows. The third does not.

This needs **no grammar change**: spec §3.8a already defines `op<pipelined, N>`, `pipelined` is already a reserved keyword, `lower_pipelined_calls` already rewrites `acc@N <= op<pipelined,N>(args)` into a staged register cascade, and `--staged-ops` already emits registry-backed modules. This is registry rows plus characterized staged implementations, per `doc/proposal_pipelined_operators.md`.

## Measurements

Yosys 0.67, one isolated operator per module, the repo's technology-independent depth flow (`synth -flatten`; `abc -fast -g AND,OR,XOR,NAND,NOR,XNOR,ANDNOT,ORNOT,MUX`; `stat`; `ltp`) — i.e. `tests/fp_v1/synth/flow.ys.tmpl`. Depth is gate levels, area is cell count.

| operator | cells | depth | vs FMA | implied stages |
|---|---|---|---|---|
| `fma` (FP32) — **the staged baseline** | 9,154 | **301** | — | 6 (shipped) |
| `scaled_dot<B4,B4>` | 3,269 | **257** | 0.85x | **~5** |
| `scaled_quantize<B4>` | 10,721 | **222** | 0.74x | **~4–5** |
| `scaled_dequantize<B4>` | 1,345 | **34** | 0.11x | **none needed** |

`B4 = ScaledVec<FP4E2M1, 8, E8M0>`. Implied stages scale the FMA's characterized 301 levels / 6 stages (~50 levels per stage); they are a starting point for characterization, not a result.

## Reading the numbers

**`scaled_dot` is the deepest block op and the one that matters.** It is the only block operator with an FMA-shaped accumulate chain — N products feeding a summation tree — and at 257 levels it sits just under a bare FMA. It is also the operation OCP MX §6.2 defines normatively, so it is what real GEMM throughput runs through.

[...]

PR #957, opened 2026-08-24
feat(fp): gate-weighted leaf pipeline scheduler + zero-call staged render (#955)

nogate-ai:
First PR of #955 — the enabling primitive for pipelining the FP sub-operators.

## Why this shape

Characterization (#955) found the MX block quantize's critical path is **84% one `arch_f32_mul`** (186 of 222 gate levels); the block-specific glue (integer max-reduce, exponent-field scale conversion, e2m1 narrow) is shallow. So the pipelining belongs on the **FP sub-operators** — which are `FpFn` SSA DAGs the staged renderer can ingest — not on the block operators, which are direct SV-string emission with loops and many calls.

This PR delivers the two primitives that enable that, both standalone-useful (they pipeline bare `mul`/`add` too):

## 1. Gate-delay-weighted leaf scheduler

Balancing a pipeline by SSA **node count** does not balance gate **delay**: `f32_mul` is 143 nodes / SSA-depth 53 but ~186 mapped gate levels, concentrated in the 48-bit multiply. `gate_weight` assigns each `Bv::Kind` a gate-depth proxy:

| kind | weight |
|---|---|
| Var / Const / Extract / Concat / ZeroExt (wiring) | 0 |
| Not / Ite / And / Or / Xor | 1 |
| Add / Sub / Cmp | W |
| Shl / Lshr | log₂W |
| **Mul** | **2W** ← the node the cut must isolate |

`node_gate_depths` computes each node's cumulative critical-path weight in one topological pass; the cut falls at even weight fractions. This is why the FMA schedule was hand-derived with measured provenance rather than generated by node count.

## 2. Zero-call `render_sv_staged`

Generalized `callee_fn: &FpFn` → `Option<&FpFn>`. Leaf ops have no nested call: the `A_` namespace is empty, `callee_starts` is all-zero, and the single-call machinery degenerates via `call_id = usize::MAX`. **The FMA path (`Some(callee)`) is byte-unchanged** — all its staged tests still pass.

## Verification — end-to-end, not by construction

`staged_leaf_f32_mul_matches_eval_bv`: derive an N=4 `f32_mul` schedule → render the staged module → run it in **iverilog** with inputs held past the pipeline fill → every output **bit-exact** to the trusted `eval_bv` oracle across **45 vectors** (inf×1, 0×inf→nan, signed, + 40 LCG-random). Skips cleanly without iverilog.

The scheduler's topological + balance properties are pinned separately for N=2..5.

- `cargo test --release --no-fail-fast`: **1226 passed**; the 5 failures are the pre-existing terminal-width miette-wrap `*_rejected*` tests, byte-identical to `origin/main` (same run on main gives the same 5 names), green in CI.
- `cargo fmt --check` clean, no clippy warnings.

## Scope

**Correctness only.** Stage *balance* → fmax is validated by yosys characterization when the `f32_mul`/`f32_add` registry rows land (next PR), where N is set by measurement and the schedule is committed as a drift-guarded `const`. The runtime `derive_leaf_schedule` leaks arrays to `'static` only in the test; production rows never leak.

Follow-ups on #955: registry rows + surface wiring (PR 2), block-operator composition (PR 3).

### DIFF
--- a/src/fp_ir.rs
+++ b/src/fp_ir.rs
@@ -917,24 +917,163 @@ impl StagedCtx<'_> {
     }
 }
 
+// ── Gate-delay-weighted leaf scheduler (arch#955) ─────────────────────────
+//
+// The FMA's staged schedule (`FMA_F32_S6_SCHEDULE`) was hand-derived with
+// measured provenance because balancing a pipeline by SSA *node count* does
+// not balance gate *delay*: one 48-bit `Mul` node is ~dozens of gate levels
+// while a `Bxor` is one. This derives a balanced cut automatically for a
+// LEAF FpFn (zero nested calls — `f32_mul`, `f32_add`, …) by weighting each
+// `Bv` node with a proxy for its mapped gate depth, computing every node's
+// cumulative critical-path weight, then cutting at even weight fractions.
+//
+// The weights are relative, not a timing model: they only have to keep the
+// deep arithmetic (Mul, Add/Sub, Cmp, shifts) from sharing a thin stage with
+// wiring (Extract/Concat/ZeroExt) and one-level gates (bitwise, Ite). The
+// emitted schedule is validated empirically against yosys per-stage depth.
+
+/// Proxy gate depth of a single `Bv` node of the given `width`, in gate
+/// levels — used only for *relative* stage balancing (see section comment).
+fn gate_weight(kind: &Kind, width: u32) -> u32 {
+    let w = width.max(1);
+    let log2 = |x: u32| 32 - (x.max(1) - 1).leading_zeros(); // ceil-ish log2
+    match kind {
+        // Wiring — no logic on the path.
+        Kind::Var(_)
+        | Kind::Const { .. }
+        | Kind::Extract { .. }
+        | Kind::Concat(_, _)
+        | Kind::ZeroExt { .. } => 0,
+        // One gate level regardless of width.
+        Kind::Not(_) => 1,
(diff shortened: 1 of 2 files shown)


nogate-ai:
## Empirical validation of the gate-weight scheduler

Rendered staged `f32_mul` at N=2..5 and characterized each through the full nangate45 buffered flow (yosys 0.67 + OpenSTA 3.1.0, `dfflibmap` + buffered ABC with the #956 `cleanup` fix, sequential STA on `clk`):

| N | area (µm²) | fmax (MHz) |
|---|---|---|
| 2 | 6,915 | 483.6 |
| 3 | 7,508 | 506.5 |
| **4** | **7,697** | **652.3** |
| 5 | 8,675 | 471.8 |

This confirms the scheduler is balancing gate delay, not node count: fmax climbs to N=4 then **regresses** at N=5. The reason is structural — the 48-bit mantissa multiply is a **single atomic `Bv::Mul` node** that cannot be split across a register boundary, so it sets a floor. N=4 optimally isolates it (its stage ≈ the multiply alone, ~1.53 ns); N=5 adds a register layer that cannot beat that floor and just costs area. A node-count scheduler would not have found this cut.

So the "stage balance validated by characterization" obligation from this PR's scope is met here: **N=4 is the characterized optimum for `f32_mul` (652 MHz)**, and it is what the block-operator composition will request.

(The staged SV was emitted via a throwaway `#[ignore]` probe, not committed.)

PR #960, opened 2026-08-25
feat(fp): pipelined block operators — scaled_quantize / scaled_dot (#955)

nogate-ai:
Adds `<pipelined, N>` staged datapaths for the two MX block operators whose critical path is dominated by FP32 multiplies (E8M0-scale blocks). Builds on the gate-weighted leaf scheduler + zero-call staged renderer from #957.

## Result

| operator | single-cycle | staged | II |
|---|---|---|---|
| `scaled_quantize<B4>` | 233 MHz | **318 MHz** | 1 |
| `scaled_dot<B4>` | 189 MHz | **298 MHz** | 1 |

Nangate45, buffered, OpenSTA. Together they pipeline the `quantize → dot` chain from **83 MHz single-cycle to ~298 MHz** at initiation interval one.

## What's here

**`scaled_quantize<Fmt, pipelined, N>(v)`** — scale-compute (comb) → register `{v, inv, scale, special}` → N staged `ArchF32MulStaged4` multiplies → comb narrow+pack. Fixed latency 5.

**`scaled_dot<pipelined, N>(a, b)`** — N exact products → `f32_add` reduction tree (one level per stage) → two power-of-two scale multiplies. Fixed latency `3 + tree-depth` (6 for an 8-element block). Reuses the generic `Name<pipelined, N>` parse — **no parser/AST change**; the operand block shape is resolved from the module's declarations at lowering time.

Both are emitted only under `arch build --staged-ops`; otherwise they lower to the combinational operator inside the pipe_reg cascade (correct latency, un-retimed). `N` is **fixed per shape** (not a free choice like `fma`); the type checker reports the required `N`.

## Verification — throughput, not stable-input

Each operator carries a **Verilator throughput lockstep** (`tests/staged_{quantize,dot}_lockstep_test.rs`): staged vs cascade, **3000 back-to-back random cycles, bit-exact, II=1**. Verilator specifically — iverilog mis-simulates the staged fp helpers' constant bit-selects (its own "sorry: all bits will be included" warning), which produces false mismatches.

Two bugs were caught by this that stable-input testing missed:
- a scale-byte pipeline **off-by-one** in the dot (tree output lands at edge `1+tree_depth`, scale-mul-1 was fed the byte one edge too late) — ~97% mismatch under throughput, invisible when inputs are held; an auto-offset comparison proved it a real datapath bug, not a latency skew;
- the staged quantize helper functions weren't emitted under `--staged-ops` (the comb call that used to pull them is gone) — fixed with a `needs_fp_lib` flag on the staged site.

The staged `f32_mul` schedule is committed as a drift-guarded const (`F32_MUL_S4_SCHEDULE`).

- `cargo test --no-fail-fast`: 1229 passed; the 5 failures are the pre-existing terminal-width `*_rejected*` diagnostic tests, byte-identical to `main`, green in CI.
- Spec §3.8a + reference card updated; skill snapshots re-synced (doc-drift).

## Scope

E8M0 scales only (boundary-compare UE4M3 is a follow-up); coarse per-level pipelining (sub-op staging toward the multiply's ~650 MHz is a later optimization).

Progress on #955.

### DIFF
--- a/doc/ARCH_HDL_Specification.md
+++ b/doc/ARCH_HDL_Specification.md
@@ -978,6 +978,18 @@ type checker reports the required `N` if a different one is written. As with
 otherwise the call lowers to the combinational operator inside the pipe_reg
 cascade (correct latency, un-retimed).
 
+`scaled_dot<pipelined, N>` additionally requires a **power-of-two block size**.
+Its coarse per-level staging pipelines the `f32_add` reduction tree one level
+per stage, which stays latency-balanced only when the tree is perfect — i.e.
+the element count is a power of two. For a non-power-of-two count the tree
+carries an odd element across a round rather than padding with zero (padding
+would turn `-0.0` into `+0.0`, a real value change), so that element crosses
+fewer register edges than its siblings and would reach the scale multiply a
+cycle early. The type checker rejects a non-power-of-two block rather than emit
+a skewed pipeline; the combinational `scaled_dot` (drop `pipelined`) handles any
+`N`. `scaled_quantize` has no such restriction — its per-element multiplies run
+in parallel at a uniform depth.
+
 **The call's depth is authoritative; the tap is a consistency check.** A
 `<pipelined, N>` call produces a value with latency `N` — it must be bound
 into a `pipe_reg<T, N>`-typed target via the matching `@N` tap, exactly like
--- a/doc/ARCH_HDL_Specification.md
+++ b/doc/ARCH_HDL_Specification.md
@@ -965,6 +965,19 @@ characterized fmax — the concrete set is intentionally kept out of this
 normative section (it churns as implementations are added) and lives in the
 generated `doc/generated/pipelined_ops.md` instead.
 
+The **block operators** of §3.11 take the same variant:
+`scaled_quantize<Fmt, pipelined, N>(v)` and `scaled_dot<pipelined, N>(a, b)`,
+for `E8M0`-scale blocks (the boundary-compare `UE4M3` scale is not pipelined
+yet). Their critical path is dominated by the per-element FP32 multiplies, so
+those are staged while the block-specific work (the shared-scale reduction and
+the narrowing) stays combinational. Unlike `fma`, `N` here is **not a free
+choice**: each block shape has one fixed staged latency (a quantize is `5`; a
+dot is `3` plus its reduction-tree depth — `6` for an 8-element block), and the
+type checker reports the required `N` if a different one is written. As with
+`fma`, the staged form is emitted only under `arch build --staged-ops`;
+otherwise the call lowers to the combinational operator inside the pipe_reg
+cascade (correct latency, un-retimed).
+
 **The call's depth is authoritative; the tap is a consistency check.** A
 `<pipelined, N>` call produces a value with latency `N` — it must be bound
 into a `pipe_reg<T, N>`-typed target via the matching `@N` tap, exactly like
--- a/doc/Arch_AI_Reference_Card.md
+++ b/doc/Arch_AI_Reference_Card.md
@@ -295,6 +295,8 @@ scaled_quantize<Fmt, policy, rounding>(v)   // policy ∈ floor_pow2|ceil_pow2|e
                                      //   default policy is scale-dependent: floor_pow2 (E8M0), exact (UE4M3)
 scaled_dequantize(b)                 // block → Vec<FP32,N> (scale applied; the ONLY way to read a lane)
 scaled_dot(a, b)                     // two blocks of the SAME type → FP32 (block dot product, FP32 accumulate)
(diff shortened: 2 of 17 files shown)


nogate-ai:
## ⛔ Blocking: `scaled_dot<pipelined, N>` silently miscompiles for every non-power-of-2 `N`

Flagged by the daily-review sweep as stranded; **do not merge until this is closed.** A code-review trace (re-verified against `origin/claude/staged-block-quantize`) found a silent wrong-answer datapath bug.

### Root cause
`sv_staged_dot` (`src/fp_block.rs`) emits the reduction tree one register layer per *level* and advances `reg_of` **only for add destinations**:

```rust
for (dst, _, _) in level {
    r{lvl}_{dst} <= c{lvl}_{dst};
    reg_of.insert(*dst, format!("r{lvl}_{dst}"));   // only add-dsts advance
}
```

`dot_schedule` carries a lone trailing value forward un-paired on any odd count (`nxt.push(cur[i])`). A carried value is **not** an add destination at the level it skips, so it is never re-registered — its `reg_of` entry stays at the shallow register while the operand it is later added to has advanced several register edges deeper. In an II=1 pipeline the two registers therefore hold products from **different input blocks**.

### Concrete trace, `N = 6`
Levels: `{6,7,8} / {9} / {10}`, with `t10 = t9 + t8`.
- `reg_of[8] = r0_8` (0-deep, re-clocked every cycle with the newest product 8)
- `reg_of[9] = r2_9` (2-deep)

At the cycle the level-3 add fires, `r0_8` holds block *k*'s product-8 while `r2_9` holds block *k−2*'s partial sum → the dot combines operands from different blocks, wrong every cycle. Same for `N = 3, 5, 6, 7, 9, 10, 12, 24, …` — i.e. every `N` that is not a power of two.

### Why the test suite misses it
- The typecheck arm accepts any `ScaledVec` with `N ≥ 1` — no power-of-2 restriction — and computes `staged_dot_binding_latency(N)` happily for `N = 6`. Valid, typechecked user code compiles under `--staged-ops` to wrong hardware with **no diagnostic**.
- `tests/staged_dot_lockstep_test.rs` exercises **only `N = 8`** (power of two → perfectly balanced → no carries → correct). It cannot catch this.

Mitigants: opt-in `--staged-ops`, E8M0-only, and canonical MX blocks are size 32 (a power of two). But this is a silent wrong-answer on documented-supported input.

### Suggested fix (either, then re-verify)
1. **Cheap/honest:** reject non-power-of-2 `N` in the `scaled_dot<pipelined>` typecheck with a clear error. *(Note: this restricts an accepted type-system input — it's a surface-visible constraint, so pair it with a one-line spec §3.8a note.)*
2. **Proper:** emit pass-through delay registers so a level-skipping operand is re-registered once per skipped level (delay-balance the tree), making all `N` correct.

Whichever fix lands, **extend `staged_dot_lockstep_test` to a non-power-of-2 `N` (e.g. 6)** — that is the regression guard this bug proves is missing. The rest of the PR (default-path gating, the `scaled_quantize` staged datapath, docs) reviewed sound.

nogate-ai:
## Non-power-of-two block-size miscompile — fixed (reject) in 2a38000

The staged `scaled_dot` had a silent miscompile for **non-power-of-two** block sizes, invisible to the N=8 lockstep test. Root cause: the coarse per-level staging pipelines the `f32_add` reduction tree one level per stage, which is latency-balanced only for a perfect (power-of-two) tree. For a non-power-of-two count, `dot_schedule` carries an odd element across a round (padding with zero is not a no-op — it turns `-0.0` into `+0.0`), so that element crosses one fewer register edge and reaches the scale multiply a cycle early.

**Reproduced structurally** with the balance checker from #968 (`tests/fp_v1/synth/pipeline_balance.py`):

| N | staged scaled_dot |
|---|---|
| 4 | BALANCED (latency 4) |
| 5 | **UNBALANCED [3, 5]** |
| 6 | **UNBALANCED [4, 5]** |
| 8 | BALANCED (latency 5) |
| 12 | **UNBALANCED [5, 6]** |

Exactly the power-of-two sizes balance. `scaled_quantize` is unaffected (parallel per-element multiplies, uniform depth — no tree).

**Fix:** the type checker now rejects `scaled_dot<pipelined, N>` on a non-power-of-two block, pointing at the combinational `scaled_dot` (which handles any N), rather than emitting a skewed pipeline. New typecheck test (accept 8; reject 6, 12); spec §3.8a + reference card updated; snapshots re-synced. Targeted suites green (`staged_dot_lockstep_test` 2/2 incl. the N=8 Verilator lockstep, `staged_quantize_lockstep_test` 1/1).

Generalizing the staging to unbalanced trees is a possible follow-up; rejecting is the sound interim. This clears the blocker noted for this PR. Once merged, the `scaled_dot` rows added to `tests/fp_v1/smt_proof/staged_ops_miter.sh` (the #968 follow-up) would guard this structurally going forward.

PR #982, opened 2026-08-30
fix(fp): lift power-of-two restriction on staged scaled_dot via delay-balancing (#980)

### LINKED ISSUE #980: Enhancement: lift the power-of-two restriction on staged scaled_dot via delay-balancing (dummy) registers
## Summary

`scaled_dot<pipelined, N>` currently **rejects non-power-of-two block sizes** at typecheck (the #960 fix): the coarse per-level staging pipelines the reduction tree one level per stage, and a non-pow2 tree carries an odd element across a level — that element crosses one fewer register and reaches the scale multiply a cycle early, a silent miscompile. Rejection was the sound interim.

The permanent fix (maintainer-directed): **insert delay-balancing pass-through registers on carried paths** instead of rejecting. Standard delay-matching.

## Why register padding is safe where zero-padding was not

`dot_schedule`'s comment correctly rejects *arithmetic* padding — pairing the odd element with an implicit `0.0` is not a no-op in FP32 (`-0.0 + 0.0 = +0.0`). **Register padding is not arithmetic padding**: a pass-through register changes *when* a value arrives, never *what* it is. Semantically invisible by construction.

```
N=6:  level1: p0+p1, p2+p3, p4+p5
      level2: t8+t9,  [t10 -> dummy reg]     <- one pass-through reg
      level3: t12 + t10_delayed              <- aligned
```

Cost: a handful of 32-bit registers, only for non-pow2 shapes.

## Why the verification stack likes it (no new machinery)

- **Lemma B passes by construction** — padding exists precisely to make min FF-depth == max FF-depth. `pipeline_balance.py`, which *caught* the #960 skew, becomes the acceptance test for the fix: the reproduction table's UNBALANCED rows (N=5 → [3,5], N=6 → [4,5], N=12 → [5,6]) must all flip to BALANCED.
- **Lemma A (UF miter, #975) is indifferent** — the extractor shorts `<=` and `=` alike, so a dummy reg is a wire in the register-shorted view; the wiring congruence is unchanged.
- **`StagedPipeline.lean` applies unchanged** — it requires only balanced feed-forward, which padding restores.
- The latency formula (`1 + tree_depth + 1 + 1`, `tree_depth = ceil(log2 N)` via `dot_levels`) already computes correctly for non-pow2 — only the emitter lacks the pass-through chains.

## Plan

1. **Emitter** (`fp_block.rs::sv_staged_dot`): track each value's production level; when a value produced at level `j` is consumed at level `k > j+1`, emit a chain of `k - j - 1` pass-through registers (reset-free, like all staged internal layers). The scale-byte delay chains (`sda*/sdb*`) already follow this pattern — generalize it to carried tree elements.
2. **Typecheck** (`typecheck.rs`): remove the power-of-two rejection (added post-#960 on the `claude/staged-block-quantize` branch).
3. **Tests**:
   - non-pow2 Verilator throughput lockstep (N=6, mirroring `staged_dot_lockstep_test.rs` — throughput, not stable-input; that's what exposed the original bug class);
   - balance rows for N ∈ {5, 6, 12} in the miter/cargo-test guards (flip the reproduction table to BALANCED);
   - the UF arithmetic miter row for one non-pow2 shape (comb reference for the same N — the comb `scaled_dot` already handles any N);
   - repurpose `staged_dot_rejects_non_power_of_two_block` into an acceptance test.
4. **Spec §3.8a + reference card + snapshots** (doc-drift CI): replace the power-of-two-restriction paragraph; the *reason* text survives, recast as "the emitter inserts delay-balancing registers on carried paths (registers delay, they do not add — unlike zero-padding, which would flip `-0.0` to `+0.0`)". User-facing restriction lift, maintainer-directed.

## Generalization note (feeds #978)

Same principle for `retime<N>`: **a short path is never a reason to reject — pad it.** Combined with the slack-register decision on #978, rejection shrinks to genuinely infeasible cases only (a *deep* path that cannot fit the stage budget even as balanced+slack).

Related: #960 (the interim rejection this lifts), #955 (block-op umbrella), #973/#975 (the balance + UF guards that gate this), #968 (StagedPipeline.lean), #978 (retime feature inheriting the pad-short-paths rule).

nogate-ai:
Closes #980.

- Emitter `fp_block.rs::sv_staged_dot` now tracks production levels and inserts `k-j-1` pass-through registers for values carried across levels (dummy regs, reset-free). Fixes silent miscompile for non-pow2 N (e.g. N=6,12) where odd element reached scale mul a cycle early.

- Typecheck `typecheck.rs` removes power-of-two rejection; now accepts any N with correct latency `1+tree_depth+1+1`.

- Tests: repurpose `staged_dot_rejects_non_power_of_two_block` to acceptance, add N=6 throughput lockstep (mirrors staged_dot_lockstep_test), add balance checks for N=5,6,12 and UF miter for N=6.

- Docs: spec §3.8a and reference card updated to describe delay-balancing (registers delay, not add).

Verified: staged_dot N=6,8 throughput lockstep and N=5,6,12 balance and N=6 UF all pass; existing N=8 behavior unchanged.

### DIFF
--- a/doc/ARCH_HDL_Specification.md
+++ b/doc/ARCH_HDL_Specification.md
@@ -978,17 +978,15 @@ type checker reports the required `N` if a different one is written. As with
 otherwise the call lowers to the combinational operator inside the pipe_reg
 cascade (correct latency, un-retimed).
 
-`scaled_dot<pipelined, N>` additionally requires a **power-of-two block size**.
-Its coarse per-level staging pipelines the `f32_add` reduction tree one level
-per stage, which stays latency-balanced only when the tree is perfect — i.e.
-the element count is a power of two. For a non-power-of-two count the tree
-carries an odd element across a round rather than padding with zero (padding
-would turn `-0.0` into `+0.0`, a real value change), so that element crosses
-fewer register edges than its siblings and would reach the scale multiply a
-cycle early. The type checker rejects a non-power-of-two block rather than emit
-a skewed pipeline; the combinational `scaled_dot` (drop `pipelined`) handles any
-`N`. `scaled_quantize` has no such restriction — its per-element multiplies run
-in parallel at a uniform depth.
+`scaled_dot<pipelined, N>` is available for any block size `N`: the coarse
+per-level staging pipelines the `f32_add` reduction tree one level per stage,
+and for a non-power-of-two count the tree carries an odd element across a
+round. The emitter inserts delay-balancing pass-through registers on carried
+paths so every element crosses the same number of stages (registers delay,
+they do not add — unlike zero-padding, which would turn `-0.0` into `+0.0`, a
+real value change). The combinational `scaled_dot` (drop `pipelined`) likewise
+handles any `N`. `scaled_quantize` has no such restriction — its per-element
+multiplies run in parallel at a uniform depth.
 
 **The call's depth is authoritative; the tap is a consistency check.** A
 `<pipelined, N>` call produces a value with latency `N` — it must be bound
--- a/doc/Arch_AI_Reference_Card.md
+++ b/doc/Arch_AI_Reference_Card.md
@@ -296,7 +296,7 @@ scaled_quantize<Fmt, policy, rounding>(v)   // policy ∈ floor_pow2|ceil_pow2|e
 scaled_dequantize(b)                 // block → Vec<FP32,N> (scale applied; the ONLY way to read a lane)
 scaled_dot(a, b)                     // two blocks of the SAME type → FP32 (block dot product, FP32 accumulate)
 scaled_quantize<Fmt, pipelined, N>(v)       // pipelined block quantize (E8M0 only); N is FIXED per shape (a quantize is 5)
-scaled_dot<pipelined, N>(a, b)              // pipelined block dot (E8M0 only); N FIXED = 3 + tree-depth (6 for N=8); block size must be a POWER OF TWO. Both need --staged-ops
+scaled_dot<pipelined, N>(a, b)              // pipelined block dot (E8M0 only); N FIXED = 3 + tree-depth (6 for N=8, 7 for N=12); any N — emitter delay-balances carried paths. Both need --staged-ops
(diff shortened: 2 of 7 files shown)


nogate-ai:
## Post-merge review — verified, with one nit

Re-verified independently on a freshly built binary (not relying on the PR's own test output), per the agent-fix-verification discipline:

**Lemma B (the check that caught the original #960 bug)** — the reproduction table flips exactly as required:

| N | pre-#982 | post-#982 |
|---|---|---|
| 3 | — | BALANCED @4 |
| 5 | UNBALANCED [3,5] | **BALANCED @5** |
| 6 | UNBALANCED [4,5] | **BALANCED @5** |
| 8 | BALANCED @5 | BALANCED @5 (unchanged) |
| 12 | UNBALANCED [5,6] | **BALANCED @6** |

**Lemma A (UF arithmetic, #975 machinery)** — staged N=6 ≡ comb N=6: `unsat`; corrupted-tree mutation: `sat` (non-vacuous). The delay chains short to wires under the extractor as designed.

**Emitted structure** — for N=6 the carried level-1 value gets exactly one pass-through (`r2_8_d <= r1_8`) before its level-3 consumption; reset-free, consistent with the staged-layer convention. `BTreeSet` determinism follow-up (8ade667) is correct.

**Tests** — all 4 lockstep + all 10 miter-file tests green, including the new `_n5/_n6/_n12` balance guards and the N=6 Verilator throughput lockstep. Spec §3.8a/reference card preserve the `-0.0` rationale while lifting the restriction. Closes #980 properly.

**Nit**: `staged_dot_rejects_non_power_of_two_block` survives as a back-compat alias that *calls the acceptance test* — a test named `rejects_X` asserting acceptance is actively confusing, and cargo test names aren't a public API; the alias can simply be deleted in any next touch of the file.

**Latent, unreachable today**: if a carried value were ever consumed at two *different* later levels, the per-level chain emission would double-declare its `_d` registers. Unreachable while `dot_schedule` builds a tree (single consumer per value); worth a one-line comment or `debug_assert` if the schedule ever generalizes.

PR #1009, opened 2026-09-07
learn: lock the store and rewrite it atomically so concurrent checks cannot wipe events.jsonl (closes #1008)

### LINKED ISSUE #1008: learn: events.jsonl rewrite in harvest_features is not concurrency-safe; concurrent arch check/build can wipe the whole store
## What happened
Five `arch check`/`arch build` processes (arch 0.72.2) ran concurrently under the same `$HOME` (a 5-seed parallel evaluation gate). Afterwards `~/.arch/learn/events.jsonl` had shrunk from 16,381,633 bytes / 10,036 events (accumulated since April) to 7,632 bytes containing only interleaved, partly corrupt feature events from those five processes. No backup existed; the store is lost.

## Root cause
`learn::harvest_features` (src/learn.rs ~L419-480) runs on every check/build of a file with doc comments. It calls `purge_features_for_files`, which does `fs::read_to_string(events.jsonl)` followed by `fs::write(events.jsonl, kept)` — a truncate-then-write with no lock and no temp-file + rename. A concurrent process that reads while another is mid-`fs::write` sees an empty or partial file and writes *that* back as the new store. `append_event` (L381) appends without a lock too, so lines also interleave.

## Reproduction (0.72.2 binary, scratch HOME)
```
T=$(mktemp -d); cp examples/<any file with /// docs>.arch $T/a.arch
HOME=$T arch check $T/a.arch                      # creates the store
python3 - <<'EOF2'   # pre-fill 3000 events (~2.4 MB)
...
EOF2
for r in 1 2 3; do for i in 1 2 3 4 5 6; do HOME=$T arch check $T/a.arch & done; wait; for i in 1 2 3; do HOME=$T arch build -o $T/a$i.sv $T/a.arch & done; wait; done
```
Observed: `events.jsonl` 2,445,081 bytes / 3003 lines before → 2,301 bytes / 3 lines after (one trial); another trial kept the size but left 8 JSON-invalid interleaved lines. Serialising the same runs behind an `flock` on `$HOME/.arch/learn/.lock` keeps the file byte-identical (3003 lines, 0 invalid), which confirms the diagnosis.

## Suggested fix
- Take an exclusive advisory lock (flock on `<learn_dir>/.lock`) around every read-modify-write and every append of `events.jsonl` (`purge_features_for_files`, `append_event`, `prune`, `learn-index` writers).
- Rewrite via temp file + atomic rename instead of `fs::write` on the live path, so readers never observe a truncated file even without the lock.
- Optionally refuse to shrink the store by more than N% in one rewrite without an explicit flag (a cheap guard against exactly this failure).

Internal (no user-facing syntax/semantics change). Found during the TASK5 benchmark regeneration on 2026-09-06.

nogate-ai:
Fixes #1008.

## Problem
`harvest_features()` (every `arch check`/`arch build` of a doc-commented file) rewrote `~/.arch/learn/events.jsonl` with truncate+write and no lock. Concurrent arch processes under one `$HOME` raced on it; on 2026-09-06 five parallel checks reduced a 16 MB / 10,036-event store to a few interleaved lines.

## Fix (internal; no language, CLI or output change)
- Exclusive advisory lock on `<learn_dir>/.lock` (`std::fs::File::lock`) around every append and read-modify-write of `events.jsonl`, `index.json`, `retrieval_counts.json`; `harvest_features` does its purge + re-append as one critical section.
- All rewrites via temp file + rename (`write_atomic`), so readers never see a truncated file. Readers stay lock-free.
- Pending failure records written atomically too.

## Verification
- `cargo test --lib learn::` (10 passed, incl. new `write_atomic_replaces_without_leaving_temp_files`, `store_lock_is_exclusive_across_handles`).
- New integration test `test_learn_store_survives_concurrent_checks_issue_1008`: 2000-event store + 24 concurrent checks → all events kept, one feature event, no interleaving, no temp files.
- Manual repro, 3 rounds × 9 concurrent runs on a 2.4 MB store: unpatched 0.72.2 → 2 KB left; this branch → byte-identical.
- `cargo fmt --check` clean; clippy adds nothing (remaining learn.rs notes are pre-existing).
- Full `cargo test --release` running as the long verification; follow-up commits if it turns anything up.

Found during the TASK5 benchmark regeneration. The benchmark keeps its 0.72.2 pin and runs the pinned binary behind an external flock wrapper; this fix is for future releases.

### DIFF
--- a/src/learn.rs
+++ b/src/learn.rs
@@ -56,6 +56,38 @@ pub fn learn_dir() -> std::io::Result<PathBuf> {
     Ok(dir)
 }
 
+/// Exclusive advisory lock over the whole store (`<learn_dir>/.lock`), held for the
+/// duration of any append or read-modify-write of `events.jsonl`, `index.json` or
+/// `retrieval_counts.json`. Released when the guard drops (the OS releases it on
+/// process exit too). Blocking, so concurrent `arch check`/`arch build` processes
+/// queue instead of racing: without this, a rewrite done as truncate+write could be
+/// read half-written by a sibling process, which then wrote that empty view back
+/// and wiped the store (issue #1008).
+struct StoreLock(#[allow(dead_code)] fs::File);
+
+fn lock_store(dir: &std::path::Path) -> std::io::Result<StoreLock> {
+    let f = fs::OpenOptions::new()
+        .create(true)
+        .read(true)
+        .write(true)
+        .truncate(false)
+        .open(dir.join(".lock"))?;
+    f.lock()?;
+    Ok(StoreLock(f))
+}
+
+/// Replace `path` atomically: write a sibling temp file, then rename it over the
+/// target, so a reader never observes a truncated or half-written file.
+fn write_atomic(path: &std::path::Path, contents: &str) -> std::io::Result<()> {
+    let tmp = path.with_file_name(format!(
+        ".{}.tmp.{}",
+        path.file_name().and_then(|n| n.to_str()).unwrap_or("store"),
+        std::process::id()
+    ));
+    fs::write(&tmp, contents)?;
+    fs::rename(&tmp, path)
+}
+
 /// Is learning capture enabled? Honors `ARCH_NO_LEARN=1` as an opt-out.
 pub fn is_enabled() -> bool {
     match std::env::var("ARCH_NO_LEARN") {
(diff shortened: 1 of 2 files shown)


nogate-ai:
Long verification done: full `cargo test --release` on 736cbdd1 — every test binary green (no failures, no panics); the only non-ok line is a pre-existing ignored test.

